PR #2, opened 2015-05-05
Added start-on-boot entry addition for debian and rpm based distros

iliapolo:


cloudify_agent/api/pm/initd.py
```diff
@@ -95,6 +96,11 @@ def _validate(file_path):
                          .format(self.config_path))
         self._create_config()
 
+        if self.start_on_boot:
+            self._create_start_on_boot_entry()
+            self.runner.sudo('sudo update-rc.d {0} defaults'
```
nir0s: sudo-ing twice...

cloudify_agent/api/pm/initd.py
```diff
@@ -280,6 +286,27 @@ def _create_config(self):
...
+                # rpm based distribution, run chkconfig
+                self.runner.sudo('/sbin/chkconfig --add {0}'
+                                 .format(self.name))
+                self.runner.sudo('/sbin/chkconfig {0} on'
+                                 .format(self.name))
+            else:
+                raise errors.DaemonConfigurationError(
+                    "Cannot create a start-on-boot entry. Unknown "
+                    "distribution base. Supported distributions bases are "
+                    "debian and RPM"
```
nir0s: should be debian and rhel

cloudify_agent/api/pm/initd.py
```diff
@@ -280,6 +286,27 @@ def _create_config(self):
         self.runner.sudo('cp {0} {1}'.format(rendered, self.config_path))
         self.runner.sudo('rm {0}'.format(rendered))
 
+    def _create_start_on_boot_entry(self):
+
+        response = self.runner.run('which dpkg', exit_on_failure=False)
+        if response.code == 0:
+            # debian based distribution, run update-rc.d
+            self.runner.sudo('update-rc.d {0} defaults'.format(self.name))
+        else:
```
nir0s: This is kind of not scalable if we decide to add more distros:
Should it be more like:

```python
if self.runner.run('which dpkg', exit_on_failure=False) == 0:
    self.runner.sudo(...)
    return
elif self.runner.run('which rpm', exit_on_failure=False) == 0:
   self.runner.sudo(...)
   return
else:
   raise ...
```

?

cloudify_agent/api/pm/initd.py
iliapolo: why? aren't there other distros that use RPM that are not RHEL where this can work on? openSuse, fedora...

cloudify_agent/api/pm/initd.py
iliapolo: yeah this looks better. ill fix

PR #21, opened 2015-08-18
CFY-2649 fix failing test_local_agent_from_package test

m1keil:
appveyor environment change caused package install under wrong python (miniconda 64bit)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,13 @@ environment:
 install:
 
   #################################
+  # Change Python Registry
+  #################################
+
+  - reg ADD HKCU\Software\Python\PythonCore\2.7\InstallPath /ve /d "C:\Python27" /t REG_SZ /f
+  - reg ADD HKLM\Software\Python\PythonCore\2.7\InstallPath /ve /d "C:\Python27" /t REG_SZ /f
+
+  #################################
   # Installing Inno Setup
   #################################
 


dankilman:
what's with wheel's wheel?

m1keil:
what do you mean?

dankilman:
I don't remember why do we need to add a wheel?
The test seemed to work fine previously without it

m1keil:
oh, right, but I think it's better to leave it there on the safe side. it
is possible python is old enough that it comes without wheel pre installed.

dankilman:
I would rather remove all binaries from this repository altogether.
Considering this will require a lot of work fixing tests that will need to download things, I didn't do it yet.
With that said, I would prefer not adding new binaries if not absolutely necessary

dankilman:
I would add to that, that I wish we hadn't added these binaries to begin with.
When we find the time, I think we should remove them and wipe the out of git's history altogether as currently, cloning this repo takes way more time then it should

m1keil:
agree

nir0s:
By the way, I already sent a pull request for removing those binaries from
the cli repo. Since then, there have been more commits of course. Once I'm
back, I can wipe both the the agent and cli repos.

m1keil:
as far as i understand, this requires massive re-clone by anyone who might using the repo. @dankilman am I correct?

dankilman:
Yes, but this will be for the greater good

dankilman:
So, are you removing this wheel?

PR #24, opened 2015-08-20
CFY-3431 removed diamond plugin dependency from setup.py

adamlavie:


nir0s:
Does that mean that the rest service and management worker no longer need
it installed?

PR #26, opened 2015-08-24
Fix path

fogelomer:


cloudify_agent/resources/pm/nssm/nssm.conf.template
```diff
@@ -19,7 +19,17 @@ if %errorlevel% neq 0 exit /b %errorlevel%
 
 echo Setting service environment
 
+set "AGENT_PATH={{ virtualenv_path }}\Scripts"
+set NEW_PATH=%AGENT_PATH%;%PATH%
+
+set NEW_PATH=%NEW_PATH:"=%
+set "NEW_PATH1="%NEW_PATH""
+
+echo %NEW_PATH1%
```
dankilman: Add some description to the `echo`
something like `echo New environment path will be %NEW_PATH1%`

cloudify_agent/tests/api/pm/__init__.py
```diff
@@ -410,6 +410,19 @@ def _check_env_var(var, expected_value):
         for key, value in expected.iteritems():
             _check_env_var(key, value)
 
+        def _check_env_path():
+            _path = self.celery.send_task(
```
dankilman: please extract this logic that is used twice

cloudify_agent/tests/api/pm/__init__.py
```diff
@@ -410,6 +410,19 @@ def _check_env_var(var, expected_value):
...
+                name='mock_plugin.tasks.get_env_variable',
+                queue=daemon.queue,
+                args=['PATH']).get(timeout=5)
+            if os.name == 'nt':
+                self.nose.tools.assert_in(
+                    str(daemon.virtualenv_path) + '\scripts', _path)
+            if os.name == 'posix':
```
dankilman: no need for `if os.name == 'posix'` a simple `else` should suffice

cloudify_agent/tests/api/pm/__init__.py
```diff
@@ -410,6 +410,19 @@ def _check_env_var(var, expected_value):
         for key, value in expected.iteritems():
             _check_env_var(key, value)
 
+        def _check_env_path():
+            _path = self.celery.send_task(
+                name='mock_plugin.tasks.get_env_variable',
+                queue=daemon.queue,
+                args=['PATH']).get(timeout=5)
+            if os.name == 'nt':
+                self.nose.tools.assert_in(
```
dankilman: you can simply do
`self.assertIn(...)` instead of `self.nose.tools.assert_in`
This will be more consistent with the way all our others asserts are written

cloudify_agent/tests/api/pm/__init__.py
```diff
@@ -410,6 +410,19 @@ def _check_env_var(var, expected_value):
...
+                    str(daemon.virtualenv_path) + '\scripts', _path)
```
dankilman: instead of `str(daemon.virtualenv_path) + '\scripts'`
please do '{0}\scripts`.format(daemon.virtualenv_path)
btw: shouldn't it be `\Scripts` with a capital s?

PR #35, opened 2015-09-24
CFY-3040 agents update

konradxyz:


dankilman:
Just merged another branch which conflict with yours (mostly around configuration) so please sort these out when you find the time.

cloudify_agent/api/utils.py
```diff
@@ -160,7 +167,7 @@ def get_home_dir(username=None):
         if username is None:
             return os.path.expanduser('~')
         else:
-            return os.path.expanduser('~{0}'.format(username))
+            return get_windows_home_dir(username)
```
dankilman: what was the problem with the previous implementation?

cloudify_agent/tests/utils.py
```diff
@@ -14,14 +14,21 @@
...
 import logging
+import platform
 import time
 import socket
 import subprocess
 import os
 import filecmp
 import tarfile
+
+import cloudify_agent
```
dankilman: this import should come after the `cloudify` imports in its own section

cloudify_agent/api/utils.py
konradxyz: It was working properly only when home directories of currently logged user and user we are asking for were in the same directory.
It is not the case (by default) when one of those users is standard local user and the other one is the user that runs services (some local admin account I believe) - which happened to be the case when trying to install new agents on windows via an old agent.

cloudify_agent/shell/commands/installer.py
```diff
@@ -0,0 +1,49 @@
...
+    os.environ[utils.internal.CLOUDIFY_DAEMON_USER_KEY] = str(
+        cloudify_agent['user'])
+    if 'basedir' not in cloudify_agent:
+        cloudify_agent['basedir'] = utils.get_home_dir(cloudify_agent['user'])
+    configuration.directory_attributes(cloudify_agent)
+    installer = prepare_local_installer(cloudify_agent)
+    installer.create_agent()
+    installer.configure_agent()
+    installer.start_agent()
+    if output_agent_file is not None:
```
dankilman: if no path is given then no configuration will be written for the new agent or is this in addition to the default location? if no, then isn't it required for valid functionality of other operations?

cloudify_agent/tests/shell/commands/test_install_local.py
```diff
@@ -0,0 +1,130 @@
...
+from cloudify_agent.installer.config import configuration
+
+from cloudify_agent.tests.installer.config import mock_context
+
+
+class TestInstaller(BaseTest):
+
+    @classmethod
+    def setUpClass(cls):
+        cls._resources_dir = tempfile.mkdtemp(
```
dankilman: BaseTest already creates a top level temp work dir (and takes care of cleaning it) so perhaps it would be better to simply create the resources dir under it?

cloudify_agent/shell/commands/installer.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@click.command()
+@click.option('--agent-file',
+              help='Path to dictionary describing agent to install.',
+              type=click.File())
+@click.option('--output-agent-file',
+              help='Path to output agent configuration')
+@handle_failures
+def install_local(agent_file, output_agent_file):
```
dankilman: can the command be `install-local` instead of `install_local` or maybe both work as well?

konradxyz: I think we should pick one - there is no point in making it more complex.
Both - and _ work for me.
Which one would you prefer?

dankilman: `-`

cloudify_agent/shell/commands/installer.py
konradxyz: This path is used in order to return value from this script - file should be immediately deleted by caller of this script.
So, even if you do not specify output_agent_file, agent will be installed properly.
You simply might not have an access to full cloudify_agent structure.

cloudify_agent/tests/shell/commands/test_install_local.py
konradxyz: Current solution creates agent package only once. if we were to use directory from BaseTest, this directory would be different for all test methods and we would either have to create agent package multiple times or use some more complex mechanism of transferring agent package between directories.
I think current solution is better.

dankilman: ok

cloudify_agent/shell/commands/installer.py
dankilman: does this mean that file under `~/.cfy-agent` will still be created?

konradxyz: Yes, by the call to installer.create_agent() in 44th line.
This parameter is completely unrelated to ~/.cfy_agent directory.

cloudify_agent/installer/config/configuration.py
```diff
@@ -162,6 +166,51 @@ def cfy_agent_attributes(cloudify_agent):
...
+
+@group('cfy-agent')
+def _add_cfy_agent_defaults(cloudify_agent):
+    pass
+
+
+def reinstallation_attributes(cloudify_agent):
+    _cfy_agent_attributes_no_defaults(cloudify_agent)
+    _add_cfy_agent_defaults(cloudify_agent)
+    if cloudify_agent.get('basedir'):
```
dankilman: when is `basedir` not configured?

konradxyz: There are two options:
1. We are installing new agent using old agent in version 3.3 - then we use basedir from cloudify_agent struct of an old agent.
2. We are installing new agent using old agent in version pre 3.3 - then we do not have basedir here. It is detected in https://github.com/codilime/cloudify-agent/blob/CFY-3040_agents_update/cloudify_agent/shell/commands/installer.py#L41
After that, we add directory_attributes.

cloudify_agent/operations.py
```diff
@@ -165,3 +182,98 @@ def _save_daemon(daemon):
...
+def _get_broker_url(ctx, agent):
+    agent = copy.deepcopy(agent)
+    broker = _get_broker_config(ctx)
+    for k, v in broker.iteritems():
+        if not agent.get(k):
+            agent[k] = v
+    return utils.internal.get_broker_url(agent)
+
+
+def _generate_new_agent_name(old_agent_name, suffix):
```
dankilman: This is a nice function. I think we should replace the current internal.utils.generate_agent_name() with something similar to this. at the very least lets replace the call to  internal.utils.generate_agent_name() in the restart operation with this call.

cloudify_agent/operations.py
```diff
@@ -165,3 +182,98 @@ def _save_daemon(daemon):
...
+            del(os.environ['CELERY_BROKER_URL'])
+        else:
+            os.environ['CELERY_BROKER_URL'] = env_broker_url
+    # Make sure that new celery agent was started:
+    agent_status = app.control.inspect(destination=[
+        'celery@{0}'.format(returned_agent['name'])])
+    if not agent_status.active():
+        raise NonRecoverableError('Could not start agent.')
+    # Setting old_cloudify_agent in order to uninstall it later.
+    ctx.instance.runtime_properties['old_cloudify_agent'] = old_agent
```
dankilman: where are we uninstalling the old agent? are we?

konradxyz: Uninstall is not available yet.
I am working on this but it will not be added in this iteration.

cloudify_agent/operations.py
konradxyz: I am not sure how can we replace generate_agent_name with this function.
I'll move this function to internal.utils so it is easily accessible and use it for both restart and create_new_agent_dict.

PR #36, opened 2015-09-29
Cfy 3224 support rabbitmq security

geokala:


cloudify_agent/api/pm/base.py
```diff
@@ -300,6 +329,37 @@ def validate_optional(self):
         self._validate_autoscale()
         self._validate_host()
 
+    def _get_broker_port(self):
+        """
+        Determines the broker port if it has not been provided. Only intended
+        to be called before self.broker_port has been set and after
+        self.broker_ssl_cert has been set.
+        """
+        if self.broker_ssl_cert == '':
```
nir0s: if not self.broker_ssl_cert

nir0s: might be nicer to add the ports to https://github.com/cloudify-cosmo/cloudify-agent/blob/master/cloudify_agent/api/defaults.py: so that they're available for both SSL and non-SSL..
and then do something like this:

```python
return defaults.BROKER_SSL_PORT if self.broker_ssl_cert else defaults.BROKER_PORT
```

cloudify_agent/api/pm/base.py
```diff
@@ -274,6 +287,22 @@ def __init__(self, logger=None, **params):
                               backend=self.broker_url)
         self._celery.conf.update(
             CELERY_TASK_RESULT_EXPIRES=defaults.CELERY_TASK_RESULT_EXPIRES)
+        if self._get_ssl_cert_path() != '':
```
nir0s: if not self._get_ssl_cert_path()

cloudify_agent/api/pm/base.py
```diff
@@ -274,6 +287,22 @@ def __init__(self, logger=None, **params):
...
+            self._celery.conf.BROKER_USE_SSL = {
+                'ca_certs': self._get_ssl_cert_path(),
+                'cert_reqs': ssl.CERT_REQUIRED,
+            }
+
+    def _create_celery_conf(self):
+        utils.render_template_to_file(
+            template_path='pm/shared/worker_conf.py.template',
+            file_path=os.path.join(self.virtualenv,
```
nir0s: we should discuss this. seems very fragile.

cloudify_agent/api/pm/base.py
```diff
@@ -300,6 +329,37 @@ def validate_optional(self):
...
+        if self.broker_ssl_cert == '':
+            return 5672
+        else:
+            return 5671
+
+    def _create_ssl_cert(self):
+        """
+        Put the broker SSL cert into a file for AMQP clients to use.
+        """
+        if self.broker_ssl_cert != '':
```
nir0s: if not self.broker_ssl_cert

cloudify_agent/resources/pm/nssm/nssm.conf.template
```diff
@@ -13,6 +13,7 @@ echo Installing the agent as a windows service...
 --pidfile={{ pid_file }} ^
 --logfile={{ log_file }} ^
 --include={{ includes }} ^
+--config {{ work_dir }}/worker_conf.py
```
nir0s: I think this will fail...
you need a caret at the end of the line...

geokala: Correction pushed.

cloudify_agent/api/pm/base.py
geokala: Agreed- at best it's a bit ugly, and at worst it could fail interestingly. The problem with both this and the dynamic creation of the cert file are that I can't currently see a safer, saner, and (importantly) working alternative.

cloudify_agent/api/pm/base.py
noakup: :+1:

cloudify_agent/api/pm/base.py
```diff
@@ -205,10 +208,16 @@ def __init__(self, logger=None, **params):
...
+        self.broker_ssl_cert = params.get(
+            'broker_ssl_cert', '')
         self.broker_ip = params.get(
             'broker_ip') or self.manager_ip
+        # Port must be determined after SSL cert has been set in order for
+        # intelligent default to work correctly
         self.broker_port = params.get(
-            'broker_port') or defaults.BROKER_PORT
+            'broker_port') or self._get_broker_port()
+        self.broker_user = params.get('broker_user', 'testuser')
```
noakup: ?

cloudify_agent/api/pm/base.py
```diff
@@ -217,10 +226,14 @@ def __init__(self, logger=None, **params):
             'name') or self._get_name_from_manager()
         self.queue = params.get(
             'queue') or self._get_queue_from_manager()
-        self.broker_url = params.get(
-            'broker_url') or defaults.BROKER_URL.format(
-            self.broker_ip,
-            self.broker_port)
+        # This is not retrieved by param as an option any more as it can make
+        # the _delete_amqp_queues function behave badly
```
noakup: what do you mean? (regarding the comment)

cloudify_agent/api/pm/base.py
noakup: I didn't understand what's happening here. let's discuss tomorrow.

cloudify_agent/api/pm/base.py
```diff
@@ -300,6 +329,37 @@ def validate_optional(self):
...
+            )
+
+    def _get_ssl_cert_path(self):
+        """
+        Determine what path the broker SSL cert should reside in.
+        """
+        if self.broker_ssl_cert == '':
+            return ''
+        else:
+            return os.path.join(self.workdir, 'broker.crt')
```
noakup: I think this should be self._get_ssl_cert_path() and not 'broker.crt'

dev-requirements.txt
```diff
@@ -1,11 +1,11 @@
-https://github.com/cloudify-cosmo/cloudify-dsl-parser/archive/master.zip
-https://github.com/cloudify-cosmo/cloudify-rest-client/archive/master.zip
-https://github.com/cloudify-cosmo/cloudify-plugins-common/archive/master.zip
-https://github.com/cloudify-cosmo/cloudify-script-plugin/archive/master.zip
+https://github.com/cloudify-cosmo/cloudify-dsl-parser/archive/3.3m5.zip
```
noakup: I am guessing you're changing that back to master..?

cloudify_agent/api/pm/base.py
geokala: If you pass the pre-created broker_url to _delete_amqp_queues, it's used by something which, at the back-end, chokes if passed something that isn't a DNS name/IP without port (you have to pass the port in separately).

Since accepting the pre-formatted URL causes breakages in situations where a username, password, and/or port are supplied, I decided it was more sensible to disallow it rather than to support tokenising it.

cloudify_agent/api/pm/base.py
geokala: It's creating the celery worker_conf.py, which is used to configure the SSL and the URL. The URL is easy to set outside of this (though if you see the comment in the template you'll note that the way we were doing it makes the credentials appear in a process listing). The SSL is a pain to set, as you really want to get the CERT_REQUIRED value in there, and trying to set it as a multiline variable in bash is doable but gets awkward fast.

As for moving it earlier in the lifecycle: I'm not sure we can- this is part of the agent install component anyway. When we're on the manager, this is easier to cope with as we can know where we will be putting agents. As soon as we start trying to deploy onto external nodes I don't think there's a better place for generating the required configs than this process manager module. The subclassed versions actually dropped some config in already (and indeed they had this call added to them).

cloudify_agent/api/pm/base.py
geokala: This is self._get_ssl_cert_path... not sure if I'm understanding your comment?

dev-requirements.txt
geokala: Without meaning to sound like a Dr Who episode: Yes, the master is coming.

cloudify_agent/api/pm/base.py
nir0s: In general, anywhere we define the SSL or non-SSL ports, if a constants/defaults file exists, we can set them there, just like in the example above.

cloudify_agent/api/pm/base.py
nir0s: We should probably have worker_conf.py as a static file within cloudify-plugins-common so that it is accessible to the rest service as well.
We can then generate the configuration file (yaml) for it which is like cloudify_broker_conf.properties and read it from within worker_conf.py. To set and get the location for the generated properties file we can use pkg_resources.

cloudify_agent/api/pm/base.py
```diff
@@ -205,10 +208,16 @@ def __init__(self, logger=None, **params):
...
         self.user = params.get('user') or getpass.getuser()
+        self.broker_ssl_cert = params.get(
+            'broker_ssl_cert', '')
         self.broker_ip = params.get(
             'broker_ip') or self.manager_ip
+        # Port must be determined after SSL cert has been set in order for
+        # intelligent default to work correctly
         self.broker_port = params.get(
-            'broker_port') or defaults.BROKER_PORT
+            'broker_port') or self._get_broker_port()
```
nir0s:
```python
self.broker_port = params.get('broker_port', self._get_broker_port())
```

cloudify_agent/api/pm/base.py
```diff
@@ -217,10 +226,14 @@ def __init__(self, logger=None, **params):
             'name') or self._get_name_from_manager()
         self.queue = params.get(
             'queue') or self._get_queue_from_manager()
-        self.broker_url = params.get(
```
nir0s: @dankilman, do we actually need broker_url?

geokala: I'm not certain. It may be used for something within the class or its descendants. Either way, as it had it before I wanted to ensure it was still present, but that it wouldn't cause problems by effectively overriding the supplied username, password, etc, which would then still be used by _delete_amqp_queues (see lines below, comment for which will be updated when I reach that stage).

dankilman:
travis is failing

cloudify_agent/api/pm/base.py
```diff
@@ -15,6 +15,8 @@
 
 import getpass
 import os
+from pkg_resources import resource_filename
```
dankilman: `pkg_resources` is 3rd party, it should come after the `from celery import Celery` import

cloudify_agent/api/pm/base.py
```diff
@@ -85,22 +91,23 @@ class Daemon(object):
         the ip address of the broker to connect to.
         defaults to the manager_ip value.
 
-    ``broker_port``
+    ``broker_ssl_enabled``:
```
dankilman: Why did you remove the `broker_port` doc?

cloudify_agent/api/pm/base.py
```diff
@@ -274,6 +295,28 @@ def __init__(self, logger=None, **params):
                               backend=self.broker_url)
         self._celery.conf.update(
             CELERY_TASK_RESULT_EXPIRES=defaults.CELERY_TASK_RESULT_EXPIRES)
+        if self.broker_ssl_enabled:
+            self._celery.conf.BROKER_USE_SSL = {
+                'ca_certs': self._get_ssl_cert_path(),
+                'cert_reqs': ssl.CERT_REQUIRED,
+            }
+
+    def _get_celery_conf_path(self):
```
dankilman: as mentioned somewhere else, the location should probably be some place let intrusive than the site-packages dir that cloudify-agent lives in

cloudify_agent/resources/pm/detach/detach.template
```diff
@@ -17,4 +17,5 @@
 --logfile={{ log_file }} \
 --include=$(cat {{ includes_path }}) \
 --detach \
--Ofair
\ No newline at end of file
+--config=cloudify.broker_conf \
```
dankilman: what is this?

cloudify_agent/resources/pm/initd/initd.conf.template
```diff
@@ -30,7 +29,7 @@ CELERYD_LOG_LEVEL={{ log_level }}
 CELERYD_PID_FILE="{{ pid_file }}"
 CELERYD_LOG_FILE="{{ log_file }}"
 CELERYD_ENV_DIR={{ virtualenv_path }}
-CELERYD_OPTS="--events -Q {{ queue }} --hostname={{ name }} --autoscale={{ max_workers }},{{ min_workers }} --maxtasksperchild=10 -Ofair --without-gossip --without-mingle"
+CELERYD_OPTS="--events -Q {{ queue }} --hostname={{ name }} --autoscale={{ max_workers }},{{ min_workers }} --maxtasksperchild=10 -Ofair --without-gossip --without-mingle --config=cloudify.broker_conf"
```
dankilman: what is this?

cloudify_agent/resources/pm/shared/broker.crt.template
```diff
@@ -0,0 +1 @@
+{{ cert }}
```
dankilman: if all you are doing is replacing a single value, then there really isn't any need of using a template in this case is there? simply write the content of `cert` into the file directly

cloudify_agent/resources/pm/initd/initd.conf.template
geokala: Celery broker configuration- specifically, this one reads from the broker_conf config file that you've commented on elsewhere.

This allows:
a) Broker SSL configuration in a sane manner (it can be exceedingly ugly to configure from exported vars).
b) Brokers that don't display the username+password (which reside in the URL) when anyone on the system runs ps.

Well, (b) would be true if I'd actually removed the hostname from this one. *making note*

geokala:
Cannot pass travis until cloudify-plugins-common is merged.

Cannot pass appveyor as it cannot download get-pip?

cloudify_agent/resources/pm/initd/initd.template
```diff
@@ -134,6 +134,9 @@ CELERYD_PID_DIR=`dirname $CELERYD_PID_FILE`
 # Extra start-stop-daemon options, like user/group.
 if [ -n "$CELERYD_CHDIR" ]; then
     DAEMON_OPTS="$DAEMON_OPTS --workdir=$CELERYD_CHDIR"
+
+    # Ensure we start in this directory if defined so that we can find our config
+    cd $CELERYD_CHDIR
```
dankilman: remove

cloudify_agent/api/pm/base.py
```diff
@@ -781,6 +862,7 @@ def create_enable_cron_script(self):
             start_command=self.start_command(),
             status_command=self.status_command()
         )
+        self._logger.debug('TTT: {sc}'.format(sc=self.start_command()))
```
dankilman: logging should probably be removed

cloudify_agent/shell/commands/daemons.py
```diff
@@ -21,6 +21,8 @@
 from cloudify_agent.api.factory import DaemonFactory
 from cloudify_agent.shell import env
 from cloudify_agent.shell.decorators import handle_failures
+from cloudify.context import BootstrapContext
```
dankilman: these two new imports are cloudify related 3rd party so they should some between to currently two existing sections

cloudify_agent/tests/api/pm/test_base.py
```diff
@@ -32,7 +32,9 @@ def setUp(self):
         self.daemon = Daemon(
             manager_ip='manager_ip',
             queue='queue',
-            name='name'
+            name='name',
+            broker_user='guest',
```
dankilman: no real need to pass this in all the tests here as this is the default value in base.py anyway

geokala: True, but this means that if someone changes that method's signature then this test will fail (rightly), so probably worth leaving in?

dankilman: cool

cloudify_agent/shell/commands/daemons.py
```diff
@@ -133,6 +167,38 @@ def create(**params):
...
+            bootstrap_agent
+        )
+
+        try:
+            attributes['broker_user'] = broker_user
+            attributes['broker_pass'] = broker_pass
+            attributes['broker_ssl_enabled'] = \
+                bootstrap_agent.broker_ssl_enabled
+            attributes['broker_ssl_cert'] = bootstrap_agent.broker_ssl_cert
+        except AttributeError:
```
dankilman: is this really necessary? I mean, this code will always run as 3.3+ so an `AttributeError` will never be raised

geokala: It might help in the case of some local testing where the bootstrap agent isn't really a bootstrap agent but instead, e.g. an empty dict.

If that's not a concern here then it should be safe to remove.

dankilman: i think we should remove it

PR #41, opened 2015-10-17
CFY-3437 fix agent configuration

dankilman:


test-requirements.txt
```diff
@@ -4,5 +4,4 @@ unittest2
 FileServer
 
 # for creating the agent package in tests
-https://github.com/cloudify-cosmo/cloudify-agent-packager/archive/3.3m6.zip
-
+https://github.com/cloudify-cosmo/cloudify-agent-packager/archive/master.zip
```
dankilman: @nir0s what should be the version here?

nir0s: 3.5.3

dankilman: shouldn't the version tool do its thing here?

PR #43, opened 2015-10-21
CFY-3463 Wagon plugins support

dankilman:


cloudify_agent/api/plugins/installer.py
```diff
@@ -34,15 +41,36 @@ def __init__(self, logger=None):
...
 
-    def install(self, source, args=''):
-
+    def install(self, plugin, blueprint_id=None):
         """
         Install the plugin to the current virtualenv.
 
-        :param source: URL to the plugin. Any pip acceptable URL is ok.
-        :param args: extra installation arguments passed to the pip command
+        :param plugin: A plugin structure as defined in the blueprint
```
nir0s: please add the `blueprint_id` param as well.

cloudify_agent/api/plugins/installer.py
```diff
@@ -34,15 +41,36 @@ def __init__(self, logger=None):
...
         """
+        plugin_id = get_plugin_id(plugin)
+        source = get_plugin_source(plugin, blueprint_id)
+        args = get_plugin_args(plugin)
+        if plugin_id:
+            self._wagon_install(plugin_id, args)
+            return plugin['package_name']
+        elif source:
+            return self._pip_install(source, args)
```
nir0s: as far as I understand, our assumption is that when users declare a plugin in the blueprint, we will look for that plugin and fallback to installation from 'source' only if a source is declared in the blueprint as well. Now let's say that a user declared a plugin and did not intend to install from source.. but they misspelled the version and thus we couldn't find the wagon. In such a care, the plugin_id will be None and we will fallback to source. I believe that in this case, we should at least log a warning stating that the required plugin could not be found and that we're falling back.

cloudify_agent/api/plugins/installer.py
```diff
@@ -34,15 +41,54 @@ def __init__(self, logger=None):
...
+                             'name: {1}'.format(plugin_id, package_name))
+            try:
+                self._wagon_install(plugin_id, args)
+            except Exception as e:
+                raise NonRecoverableError('Failed installing managed '
+                                          'plugin: {0} [{1}][{2}]'
+                                          .format(plugin_id, plugin, e))
+            return package_name
+        elif source:
+            self.logger.info('Installing plugin from source')
```
nir0s: Please specify the source in the message.

cloudify_agent/api/plugins/installer.py
```diff
@@ -221,3 +262,89 @@ def extract_package_name(package_dir):
...
+        return None
+
+    plugin_result = max(plugins,
+                        key=lambda plug: LooseVersion(plug.package_version))
+    return plugin_result.id
+
+
+def get_plugin_source(plugin, blueprint_id=None):
+
+    source = plugin.get('source') or ''
```
nir0s: `get('source', '')` ?

cloudify_agent/api/plugins/installer.py
```diff
@@ -34,15 +41,54 @@ def __init__(self, logger=None):
...
-        :param source: URL to the plugin. Any pip acceptable URL is ok.
-        :param args: extra installation arguments passed to the pip command
+        :param plugin: A plugin structure as defined in the blueprint.
+        :param blueprint_id: The blueprint id associated with this
+                             installation. if specified, will be used
+                             when downloading plugins that were included
+                             as part of the blueprint itself.
         """
+        plugin_id = get_plugin_id(plugin)
+        source = get_plugin_source(plugin, blueprint_id)
```
nir0s: is there a reason for calling this method even if `plugin_id` is set?

cloudify_agent/api/plugins/installer.py
```diff
@@ -221,3 +262,89 @@ def extract_package_name(package_dir):
...
+        blueprint_plugins_url = '{0}/{1}/plugins'.format(
+            blueprints_root, blueprint_id)
+
+        source = '{0}/{1}.zip'.format(blueprint_plugins_url, source)
+
+    return source
+
+
+def get_plugin_args(plugin):
+    args = plugin.get('install_arguments') or ''
```
nir0s: plugin.get('install_arguments', '' )...

cloudify_agent/api/plugins/installer.py
```diff
@@ -34,15 +41,54 @@ def __init__(self, logger=None):
...
+        if plugin_id:
+            package_name = plugin['package_name']
+            self.logger.info('Installing managed plugin: {0} based on package_'
+                             'name: {1}'.format(plugin_id, package_name))
+            try:
+                self._wagon_install(plugin_id, args)
+            except Exception as e:
+                raise NonRecoverableError('Failed installing managed '
+                                          'plugin: {0} [{1}][{2}]'
+                                          .format(plugin_id, plugin, e))
```
adamlavie: str(e)

cloudify_agent/api/plugins/installer.py
```diff
@@ -221,3 +262,89 @@ def extract_package_name(package_dir):
...
+    source = plugin.get('source') or ''
+    if not source:
+        return None
+    source = source.strip()
+
+    # validate source url
+    if '://' in source:
+        split = source.split('://')
+        schema = split[0]
+        if schema not in ['http', 'https']:
```
nir0s: btw, I dunno how exactly it works with banks and the likes.. but I'm pretty sure we need to start supporting ftp for downloads :)

cloudify_agent/api/plugins/installer.py
```diff
@@ -221,3 +262,89 @@ def extract_package_name(package_dir):
         cwd=package_dir
     ).std_out
     return plugin_name
+
+
+def get_plugin_id(plugin):
+    package_name = plugin.get('package_name')
```
nir0s: Is there currently a way for us to simply provide a url/blueprint-relative-path to a wagon archive for installation?
It would seem good to me to allow users to override the querying mechanism and just allow them to state: "this is the wagon I want for this plugin and I don't give a sh*t about anything else" :)

cloudify_agent/api/plugins/installer.py
dankilman: no need for str(e) within `format`. it gets called by `format`

cloudify_agent/api/plugins/installer.py
dankilman: this is not my code. i would rather leave it as is for now as I don't want do deal with the implications of changing this now

cloudify_agent/api/plugins/installer.py
dankilman: again, copy/pasted from a different location. will not deal with this now

cloudify_agent/api/plugins/installer.py
dankilman: again

cloudify_agent/api/plugins/installer.py
dankilman: good suggestion, not sure how to expose it though

adamlavie: maybe plugin_id?

PR #47, opened 2015-10-26
CFY-3436 fix download path

dankilman:


cloudify_agent/api/plugins/installer.py
```diff
@@ -75,8 +75,8 @@ def install(self, plugin, blueprint_id=None):
 
     def _wagon_install(self, plugin_id, args):
         client = get_rest_client()
-        fd, wagon_path = tempfile.mkstemp()
-        os.close(fd)
+        wagon_dir = tempfile.mkdtemp(prefix='{0}-'.format(plugin_id))
+        wagon_path = os.path.join(wagon_dir, 'wagon.tar.gz')
```
adamlavie: plugin_dir/plugin_path/plugin.tar.gz would make more sense to me

PR #52, opened 2015-10-28
CFY-3040 broker credentials bugfix

konradxyz:


cloudify_agent/operations.py
```diff
@@ -195,17 +195,14 @@ def _get_broker_url(ctx, agent):
...
-                        'manager_port', 'broker_url', 'broker_port',
-                        'broker_ip', 'key', 'version']
-    for field in fields_to_delete:
-        if field in new_agent:
-            del(new_agent[field])
+    new_agent = {}
     new_agent['name'] = utils.internal.generate_new_agent_name(
         old_agent['name'])
     new_agent['remote_execution'] = True
+    fields_to_copy = ['windows', 'ip', 'basedir', 'user']
```
dankilman: can you explain the reason for this fix?

dankilman:
can you explain the reason for this fix?

cloudify_agent/operations.py
dankilman: when do we actually need to preserve any old property? if we simply call `reinstallation_attributes` with an empty initial new_agent dict, should that suffice?
The runtime property gets populated in the first place from similar methods in the configuration module.

konradxyz: Not exactly.
reinstallation_attributes is working differently.
Your usual configuration function first adds data from runtime properties, then properties, then defaults and if everything else fails code of the function is expected to fill in the gaps.
reinstallation_attributes on the other hand gives priority to the function code, and uses runtime properties/properties as fallback.
Otherwise we would end up installing agent in directory where old agent was installed.

Quick analysis of the code suggests that it would be sufficient to copy `windows` field.

dankilman: ok.
can you explain why need to keep the `windows` field?

konradxyz: Otherwise this line will fail:
https://github.com/cloudify-cosmo/cloudify-agent/blob/master/cloudify_agent/installer/config/configuration.py#L145
The more I think about it the more shaky it seems.
I think it might be a good idea to change this to use standard priorities from configuration module, just skipping runtime properties.

PR #69, opened 2015-11-18
Added agents alive after reboot test

konradxyz:


dankilman:
Due to differences in the way this is implemented in ubuntu and centos, please add another test for centos

system_tests/resources/reboot-vm-blueprint/scripts/application_delete.py
```diff
@@ -0,0 +1,4 @@
+from cloudify import ctx
```
dankilman: can you update this operation so that it updates a runtime property of this node instance and then at the end of `_test_agent_alive_after_reboot`, call `self.client.node_instances.list(node_id='application', deployment_id=self.test_id)[0].runtime_properties` and add an assertion of that runtime property value? Currently we are only asserting that the agent is up but we are not validating that it is actually functioning properly. Uninstall will ignore any errors it encounters, so it will always succeed even if this operation never executed.

PR #74, opened 2015-12-04
Cfy 4381 certificates support

konradxyz:


dankilman:
Do we know why appveyor is failing?

cloudify_agent/operations.py
```diff
@@ -208,27 +184,57 @@ def create_new_agent_dict(old_agent):
     new_agent['manager_file_server_url'] = get_manager_file_server_url()
     # Assuming that if there is no version info in the agent then
```
dankilman: This comment should probably be moved to where version is currently assigned.

cloudify_agent/operations.py
```diff
@@ -208,27 +184,57 @@ def create_new_agent_dict(old_agent):
...
+    # We can't pass broker_url to Celery constructor because it would
+    # be overriden by the value from broker_config.py.
+    config = {
+        'BROKER_URL': broker_url,
+        'CELERY_RESULT_BACKEND': broker_url
+    }
+    if agent['version'] != '3.2':
+        config['CELERY_TASK_RESULT_EXPIRES'] = \
+            defaults.CELERY_TASK_RESULT_EXPIRES
+    fd, cert_path = tempfile.mkstemp()
```
dankilman: Do these two lines get tested on windows?
I recall vaguely that this approach caused problems in windows once when I tried it.

cloudify_agent/operations.py
```diff
@@ -208,27 +184,57 @@ def create_new_agent_dict(old_agent):
...
+    finally:
+        os.remove(cert_path)
+
+
 def create_agent_from_old_agent(install_agent_timeout=300):
     if 'cloudify_agent' not in ctx.instance.runtime_properties:
         raise NonRecoverableError(
             'cloudify_agent key not available in runtime_properties')
     old_agent = ctx.instance.runtime_properties['cloudify_agent']
+    old_agent['version'] = old_agent.get('version', _VERSION)
```
dankilman: Can we use a mechanism similar to
https://github.com/cloudify-cosmo/cloudify-manager/pull/365/files#diff-65e4c2d1d79b4b1b3b988c9be1803305R306
instead of hard coding the default `_VERSION` value?

cloudify_agent/operations.py
```diff
@@ -208,27 +184,57 @@ def create_new_agent_dict(old_agent):
...
     new_agent = create_new_agent_dict(old_agent)
-    # We retrieve broker url from old agent in order to support
-    # cases when old agent is not connected to current rabbit server.
-    broker_url = _get_broker_url(ctx, old_agent)
-    env_broker_url = os.environ.get('CELERY_BROKER_URL')
```
dankilman: why do we no longer need the `CELERY_BROKER_URL` env variable workaround?

konradxyz:
Not really.

I've run failing test manually and I wasn't able to reproduce this error.
I've created dummy empty PR to make sure that this error is related to this PR.

cloudify_agent/operations.py
konradxyz: Yes, these lines are tested in test_operations.py:TestCreateAgentAmqp.test_create_agent_from_old_agent

cloudify_agent/operations.py
konradxyz: We no longer use this variable.
It was removed when broker_config.py was added.

cloudify_agent/operations.py
konradxyz: And of course these lines are not expected to be run in windows environment.

cloudify_agent/operations.py
konradxyz: This is good idea but I think this change is not related to problem we are solving in this PR.

cloudify_agent/operations.py
dankilman: yup, true dat.

cloudify_agent/operations.py
dankilman: can we still do it?

pleszcz:
@dankilman please close

PR #75, opened 2015-12-09
Cfy 4382 sanity checks

konradxyz:


cloudify_agent/operations.py
```diff
@@ -206,54 +183,157 @@ def create_new_agent_dict(old_agent):
...
+        raise NonRecoverableError(
+            ('Task script_runner.tasks.run is not available in agent '
+             '{0}').format(name))
+
+
+def _run_install_script(old_agent, timeout, validate_only=False):
+    # Assuming that if there is no version info in the agent then
+    # this agent was installed by current manager.
+    old_agent = copy.deepcopy(old_agent)
+    old_agent['version'] = old_agent.get('version', _VERSION)
```
dankilman: please use a mechanism similar to 
https://github.com/cloudify-cosmo/cloudify-manager/blob/master/workflows/cloudify_system_workflows/snapshot.py#L166
to extract the version instead of using a hard coded value. This change is unrelated but leaving this code as it is is very problematic

pleszcz: Could you give me one more hint how to find this version value when it's absent? Using the method you suggested doesn't work since we can't create rest client using get_rest_client from cloudify.manager (I suppose the cloudify-manager doesn't start for this test):
Traceback (most recent call last):
  File "/vagrant_data/cloudify-plugins-common/cloudify/decorators.py", line 118, in wrapper
    result = func(*args, **kwargs)
  File "/vagrant_data/cloudify-agent/cloudify_agent/operations.py", line 346, in validate_agent_amqp
    raise NonRecoverableError(result['agent_alive_crossbroker_error'])
NonRecoverableError: 'MANAGER_REST_PORT'

dankilman: I suppose you can only try reading this value if it is not already provided in `old_agent`. So, in a real manager, it should work and in a test it will never reach code that tries to instantiate a REST client

pleszcz: This is exactly what I did. However there is still the case (in test_install_new_agent) when we have a old_agent without version set and we need to provide it somehow.

dankilman:
If this is only for the test, I will try mocking it. If I recall correctly, this test runs a local workflow, so patching you do before running the test will apply.
Try mocking the `get_rest_client` attribute you import there so that it returns some mock client

cloudify_agent/operations.py
dankilman: If this is only for the test, I will try mocking it. If I recall correctly, this test runs a local workflow, so patching you do before running the test will apply.
Try mocking the `get_rest_client` attribute you import there so that it returns some mock client

pleszcz:
@dankilman Please close. New one #80

PR #80, opened 2015-12-18
CFY-4381 and CFY-4382

pleszcz:
* CFY-4381 Certificates support in install_new_agents
* CFY-4382 Sanity checks for agents

This the replacement for #75

### DIFF
--- a/cloudify_agent/operations.py
+++ b/cloudify_agent/operations.py
@@ -13,20 +13,25 @@
 #  * See the License for the specific language governing permissions and
 #  * limitations under the License.
 
+import tempfile
 import time
 import threading
+import ssl
 import sys
 import os
 import copy
+
+from contextlib import contextmanager
 
 import celery
 
 from cloudify import ctx
 from cloudify.exceptions import NonRecoverableError
 
+import cloudify.manager
+
 from cloudify.utils import get_manager_file_server_url
 from cloudify.decorators import operation
-from cloudify import utils as cloudify_utils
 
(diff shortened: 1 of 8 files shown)


cloudify_agent/operations.py
```diff
@@ -37,6 +40,8 @@
 
 from cloudify_agent.installer.config import configuration
 
+import cloudify.manager
```
dankilman: move this import so that it is right before the `cloudify.utils` import

dev-requirements.txt
```diff
@@ -1,6 +1,6 @@
 https://github.com/cloudify-cosmo/cloudify-dsl-parser/archive/3.3.1.zip
 https://github.com/cloudify-cosmo/cloudify-rest-client/archive/3.3.1.zip
-https://github.com/cloudify-cosmo/cloudify-plugins-common/archive/3.3.1.zip
+https://github.com/codilime/cloudify-plugins-common/archive/CFY-4381_certificates_support.zip
```
dankilman: please fix this path to `CFY-4381_and_CFY-4382`

cloudify_agent/tests/resources/blueprints/install-new-agent/install-new-agent.yaml
```diff
@@ -0,0 +1,83 @@
+tosca_definitions_version: cloudify_dsl_1_2
```
dankilman: can you rename this blueprint so that it ends with `-blueprint.yaml`

PR #86, opened 2016-01-03
CFY-4099 added atfork() to prevent crypto PID check failed assertion …

nirbir:
…error


funkyHat:
This was merged in #92

PR #107, opened 2016-03-25
CFY-5111 use virtualenv's method of fixing shebangs

tehasdf:
the previous _fix_virtualenv method left the scripts broken when the
virtualenv path was longer than about 256 bytes (can't run a program
with a shebang longer than 256).

virtualenv's own method of making venvs relocatable does almost
what we want (uses a /usr/bin/env shebang and a "activate this"
snippet), but is too strict (only works on not-yet-moved virtualenvs,
while agent packages have already been moved).

This changeset adapts virtualenv's method, turning off the checking.

(while very long virtualenv names might be rare/nonexistent in practice,
some system tests have a test_id long enough to break because of this;
this fixes cosmo_tester.test_suites.test_blueprints.puppet_plugin_test
.PuppetPluginStandaloneTest.test_puppet_standalone_with_resource)

### DIFF
--- a/cloudify_agent/shell/commands/configure.py
+++ b/cloudify_agent/shell/commands/configure.py
@@ -14,10 +14,14 @@
 #  * limitations under the License.
 
 import os
+import sys
+
 import click
 
+from virtualenv import (OK_ABS_SCRIPTS, is_win, path_locations,
+                        fixup_pth_and_egg_link, relative_script)
+
 from cloudify.utils import LocalCommandRunner
-from cloudify.exceptions import CommandExecutionException
 
 from cloudify_agent import VIRTUALENV
 from cloudify_agent.api import utils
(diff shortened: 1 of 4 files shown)


cloudify_agent/shell/commands/configure.py
```diff
@@ -73,53 +77,100 @@ def _disable_requiretty():
...
+    of a /usr/bin/env shebang, and a python snippet that runs the `activate`
+    script.
+    """
     from cloudify_agent.shell.main import get_logger
     logger = get_logger()
 
-    bin_dir = '{0}/bin'.format(VIRTUALENV)
+    new_shebang = _get_relative_shebang()
+    for filename in _find_scripts_to_fix(bin_dir):
+        logger.info('Making script %s relative' % filename)
```
dankilman: i think debug logging is more appropriate here

dankilman: if we already went the extra mile of copying text from virtualenv, lets modify the string formatting code so that is uses '{0} some test'.format(some_arg) so that is it more consistent with how we generally do string formatting

cloudify_agent/shell/commands/configure.py
```diff
@@ -73,53 +77,100 @@ def _disable_requiretty():
...
+            except UnicodeDecodeError:
+                # This is probably a binary program instead
+                # of a script, so just ignore it.
+                continue
+
+        if not lines:
+            continue
+
+        shebang = lines[0]
+        if not (shebang.startswith('#!') and shebang.endswith('python')):
```
dankilman: i'm thinking maybe also support ends with: python2, python2.6, python2.7, etc.. although I'm not that sure about this. our agent packages don't face this issue anyway.

tehasdf: yes, eh, there's this tradeoff here, between being more general, and needing to assume more things :D

well, the number of possibilities there is very limited so maybe let's not overthink this: i think just checking `"bin/python" in path` would cover all cases while not causing any false positives in realistic scripts?

tehasdf:
@dankilman thanks, i updated the commit to implement these changes

dankilman:
PR looks good. There is a test failing in appveyor which may be related to this change. I've re-run the build in appveyor to see if it persists. If it does, please have a look at it.

tehasdf:
Ugh, circle CI doesn't pass because of installation issues, I didn't even touch setup.py or reqs.txt :D Weird. Appveyor seems to work, though.
However! There's no tests here that would cover this very issue. I'm going to add one in the next commit, and if it breaks, I'll fix it (on monday probably). Otherwise I'll squash the commits.

setup.py
```diff
@@ -27,7 +27,8 @@
     'fabric==1.8.3',
     'wagon==0.3.0',
     'fasteners==0.13.0',
-    'pyzmq==15.1.0'
+    'pyzmq==15.1.0',
+    'virtualenv==15.0.1'
```
dankilman: Considering that `wagon` has this dependency: `"virtualenv>=12.1"`, perhaps it'd be better to have the same range here instead of the hard coded `15.0.1`

cloudify_agent/tests/installer/test_operations.py
```diff
@@ -95,6 +95,40 @@ def test_local_agent_from_package(self, _):
         env.execute('uninstall', task_retries=1)
         self.wait_for_daemon_dead(name=agent_name)
 
+    @only_os('posix')
```
dankilman: just verifying this test fails without the agent fix

tehasdf: It should, we can rerun it on CI without, but for now, here's the logs from my machine with and without the fix https://gist.github.com/tehasdf/4329cec872067c661d6e226a1eca1c2e

setup.py
tehasdf: right, changed, thanks

PR #108, opened 2016-03-30
CFY-5111 installing agent from source: make venv relocatable

tehasdf:
This is only needed for long venv paths.

### DIFF
--- a/cloudify_agent/installer/__init__.py
+++ b/cloudify_agent/installer/__init__.py
@@ -110,6 +110,14 @@ class AgentInstaller(object):
         self.runner.run('{0} install {1}'
                         .format(pip_path, source_url))
 
+        # scripts inside the virtualenv will have /path/to/venv/bin/python
+        # as their shebang. If this exceeds 128 bytes, the scripts will
+        # become non-executable, unless make the virtualenv relocatable
+        # Do this after installing the agent, so that the agent script
+        # is also made relocatable
+        self.runner.run('virtualenv --relocatable {0}'.format(
+            self.cloudify_agent['envdir']))
+
     def _from_package(self):
 
         self.logger.info('Downloading Agent Package from {0}'.format(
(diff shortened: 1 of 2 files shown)


cloudify_agent/installer/__init__.py
```diff
@@ -95,6 +95,13 @@ def _from_source(self):
         self.runner.run('virtualenv {0}'.format(
             self.cloudify_agent['envdir']))
 
+        if len(self.cloudify_agent['envdir']) > 110:
```
dankilman: 110 feels quite arbitrary. In general I agree with the comment made on JIRA that relocatable should always be applied. My only concern here is with this code's behavior on windows. Let's remove the condition altogether, if the source installation test passes on appveyor, great, otherwise, let's apply this change on linux only

cloudify_agent/tests/installer/test_operations.py
```diff
@@ -152,3 +153,36 @@ def test_3_2_backwards(self, _):
...
+
+        inputs = {
+            'source_url': self.source_url,
+            'requirements_file': self.requirements_file,
+            'name': agent_name,
+            'queue': agent_queue
+        }
+
+        blueprint_path = resources.get_resource(
+            'blueprints/3_2-agent-from-source/3_2-agent-from-source.yaml')
```
dankilman: not sure we want the 3_2 agent blueprint, the 3.2 blueprint is mainly for backwards compatibility testing.  `blueprints/agent-from-source/local-agent-blueprint.yaml` is probably more suitable here

appveyor.yml
```diff
@@ -1,7 +1,7 @@
 environment:
 
   TOX_ENV: pywin
-
+  APPVEYOR: true
```
tehasdf: hmm, sorry, I was confused - I thought appveyor was skipping some tests and that it was because of the `@only_ci`, because it didn't provide an APPVEYOR env var
but now i tried it, and it does :)
so, removed

cloudify_agent/tests/installer/test_operations.py
tehasdf: right, thanks

tehasdf:
this changeset was squashed into #107

PR #62, opened 2015-11-09
CFY-3691 Make all internal REST calls secure

noakup:
Conflicts:
	dev-requirements.txt
	packaging/linux/packager.yaml

PR #114, opened 2016-05-10
CFY-5157 Pass the security context when installing agents (DONT MERGE YET)

tehasdf:
When doing `cfy agents install`, also pass the security context
to the new agent installer.


tehasdf:
needs #62 (and probably a rebase after that is merged)

PR #131, opened 2016-07-13
CFY-5598 Use REST credentials from the context

tehasdf:


tehasdf:
This requires cloudify-cosmo/cloudify-plugins-common#248

Similar to there, the main point is in the `get_rest_client` function, which now uses the token. The token will be in the context, which is done by the plugins-common changes.
Most of the changes here have to do with actually removing the credentials so they're not stored on disk anymore.

cloudify_agent/api/pm/base.py
```diff
@@ -255,8 +259,9 @@ def __init__(self, logger=None, **params):
         self.rest_protocol = params.get(
             'rest_protocol') or defaults.REST_PROTOCOL
         self.security_enabled = params.get('security_enabled')
-        self.rest_username = params.get('rest_username')
-        self.rest_password = params.get('rest_password')
+        self._rest_username = params.get('rest_username')
```
dankilman: making these properties "hidden" (`_`) is inconsistent with the other properties

tehasdf: they need to not be stored when the daemon is serialized.

Instead, i can not store them at all, but create the rest client as a local var in __init__ and pass it to the methods that need it (that happen to only be called from __init__)

dankilman: oh, but the will make it no serialized so is get it. maybe it is worth adding a comment before these properties

cloudify_agent/api/pm/base.py
```diff
@@ -235,8 +239,23 @@ def __init__(self, logger=None, **params):
...
+            'rest_protocol') or defaults.REST_PROTOCOL
+        self.verify_rest_certificate = params.get('verify_rest_certificate')
+        self.local_rest_cert_file = params.get('local_rest_cert_file', '')
+        self.rest_cert_content = params.get('rest_ssl_cert_content', '')
+        self.security_enabled = params.get('security_enabled')
+        # REST credentials need to be prefixed with _ so they're not stored
+        # when the daemon is serialized
+        self._rest_username = params.get('rest_username')
+        self._rest_password = params.get('rest_password')
+        self._rest_token = params.get('rest_token')
```
tehasdf: all the REST parameters must be in place before `._get_name_from_manager` because that might attempt to use the REST client

cloudify_agent/tests/installer/config/__init__.py
```diff
@@ -39,5 +39,6 @@ def mock_context(agent_properties=None,
         bootstrap_context=BootstrapContext(
             bootstrap_context={'cloudify_agent': agent_context})
     )
-    setattr(context, 'installer', MagicMock())
+    context.installer = MagicMock()
+    context._get_current_object = lambda: context
```
tehasdf: btw `_get_current_object` is actually public api; pocoo projects commonly use _ to prefix attributes to avoid name conflicts, if the object is allowed to have a user-provided attribute named the same: http://werkzeug.pocoo.org/docs/0.11/local/#werkzeug.local.LocalProxy._get_current_object
(we use `proxy_tools` which is copied from werkzeug)

PR #141, opened 2016-07-27
using cert file on windows as well if found

noakup:


cloudify_agent/installer/__init__.py
```diff
@@ -341,12 +341,13 @@ def create_custom_env_file_on_target(self, environment):
...
     def download(self, url, destination=None):
+        local_cert_file = self.cloudify_agent.get('local_rest_cert_file')
         if self.cloudify_agent['windows']:
-            return self.runner.download(url,
-                                        output_path=destination,
-                                        skip_verification=True)
+            return self.runner.download(
+                url,
+                output_path=destination,
+                skip_verification=not bool(local_cert_file))
```
tehasdf: maybe use `cloudify_agent['verify_rest_certificate']`?

cloudify_agent/installer/__init__.py
```diff
@@ -341,16 +342,21 @@ def create_custom_env_file_on_target(self, environment):
             return None
 
     def download(self, url, destination=None):
+        local_cert_file = self.cloudify_agent.get('local_rest_cert_file')
+        verify_cert = self.cloudify_agent.get('verify_rest_certificate')
+        if verify_cert is None:
```
dankilman: 1) Why are we verifying this value? should we simply default to `True`?
2) If the reason to that turns out to be no, then this is certainly not the place to put such validation. The `configuration.py` module is much more suited for such tasks.

cloudify_agent/installer/__init__.py
```diff
@@ -341,16 +342,21 @@ def create_custom_env_file_on_target(self, environment):
...
+                url,
+                output_path=destination,
+                skip_verification=not verify_cert)
         else:
-            local_cert_file = self.cloudify_agent.get('local_rest_cert_file')
             return self.runner.download(
                 url,
                 output_path=destination,
-                skip_verification=not bool(local_cert_file),
+                skip_verification=not verify_cert,
```
dankilman: why didn't we use this value previously?

noakup: yes, Idan asked me to accept lukasz's comment, which makes it less relying on the creation of the cert file and clearer

cloudify_agent/installer/__init__.py
noakup: The default is actually False and we take care of it here: https://github.com/cloudify-cosmo/cloudify-agent/blob/master/cloudify_agent/shell/commands/daemons.py#L461

dankilman: ok, so please remove the validation code and use `[]` in both values instead of `get`

noakup: according to [daemons.py](https://github.com/cloudify-cosmo/cloudify-agent/blob/windows-agent-cert/cloudify_agent/shell/commands/daemons.py#L473) agent['local_rest_cert_file'] might not be set, so I think we should use `get` for it

dankilman: but according to 
https://github.com/cloudify-cosmo/cloudify-agent/blob/windows-agent-cert/cloudify_agent/installer/config/configuration.py#L207l
this is always set. configuration runs before download, daemons runs after

noakup: ok, accepted

dankilman: and by the way according to daemons.py it is always set which is why `[]` is used instead of `get`

cloudify_agent/installer/__init__.py
```diff
@@ -341,16 +341,18 @@ def create_custom_env_file_on_target(self, environment):
             return None
 
     def download(self, url, destination=None):
+        local_cert_file = self.cloudify_agent['local_rest_cert_file']
```
dankilman: i have now found the `installer.download` is used in places where an agent package isn't installed.
i think this method should get these as parameters where verify will be `True` by default and only in
https://github.com/cloudify-cosmo/cloudify-agent/blob/master/cloudify_agent/installer/__init__.py#L127
should we supply values derived from `cloudify_agent`

PR #146, opened 2016-07-29
CFY-5135 Upload cert to the agent host before downloading agent package

tehasdf:


cloudify_agent/installer/__init__.py
```diff
@@ -353,7 +370,24 @@ def download(self, url, destination=None):
...
+    def _get_certificate_file(self):
+        cert_file = self.cloudify_agent['local_rest_cert_file']
+        cert_content = self.cloudify_agent.get('rest_cert_content')
+
+        if not cert_content:
+            return None
+
+        if not self.runner.exists(cert_file):
+            with tempfile.NamedTemporaryFile(delete=False) as f:
+                f.write(cert_content.decode('string-escape'))
```
tehasdf: same as above

cloudify_agent/installer/__init__.py
```diff
@@ -329,6 +332,20 @@ def create_custom_env_file_on_target(self, environment):
...
+        cert_file = self.cloudify_agent['local_rest_cert_file']
+
+        if not os.path.exists(cert_file):
+            cert_content = self.cloudify_agent.get('rest_cert_content')
+
+            if not cert_content:
+                return None
+
+            with open(cert_file) as f:
+                f.write(cert_content.decode('string-escape'))
```
tehasdf: why is it escaped in the first place?

cloudify_agent/installer/__init__.py
```diff
@@ -259,7 +259,7 @@ def install_pip(self):
         get_pip_url = 'https://bootstrap.pypa.io/get-pip.py'
         self.logger.info('Downloading get-pip from {0}'.format(get_pip_url))
         destination = '{0}\\get-pip.py'.format(self.cloudify_agent['basedir'])
-        get_pip = self.download(get_pip_url, destination)
+        get_pip = self.runner.download(get_pip_url, destination)
```
tehasdf: we need to skip our internal cert mechanism because this just uses an external pypa.io url

cloudify_agent/installer/__init__.py
```diff
@@ -292,7 +292,7 @@ def cfy_agent_path(self):
     def install_pip(self):
         get_pip_url = 'https://bootstrap.pypa.io/get-pip.py'
         self.logger.info('Downloading get-pip from {0}'.format(get_pip_url))
-        get_pip = self.download(get_pip_url)
+        get_pip = self.runner.download(get_pip_url)
```
tehasdf: as above

cloudify_agent/installer/runners/fabric_runner.py
```diff
@@ -389,7 +389,7 @@ def download(self, url, output_path=None, skip_verification=False,
             if skip_verification:
                 args += ' --no-check-certificate'
             if certificate_file:
-                args += ' --ca-certificate={0}'.format(certificate_file)
+                args += ' --ca-certificate {0}'.format(certificate_file)
```
tehasdf: allows shell expansion of ~

cloudify_agent/installer/__init__.py
dankilman: this has something to do with the way we propagate the certificate content from the manager blueprint. specifically, we pass this as an environment variable to the management worker systemd environment configuration file and it is escaped there. @noakup can probably elaborate.

dankilman: what exactly does `string-escape` do? can you provide a snippet of how the cert looks before and after decode?

cloudify_agent/installer/__init__.py
```diff
@@ -353,7 +367,24 @@ def download(self, url, destination=None):
...
+        cert_content = self.cloudify_agent.get('rest_cert_content')
+
+        if not cert_content:
+            return None
+
+        if not self.runner.exists(cert_file):
+            with tempfile.NamedTemporaryFile(delete=False) as f:
+                f.write(cert_content.decode('string-escape'))
+            try:
+                self.runner.put_file(f.name, cert_file)
```
dankilman: The implicitness of copying a file during a simple `get` call somewhat bothers me. It will mostly surprise people that read this code. We can probably explicitly perform this step in `create_agent` and have `_get_certificate_file` only return a path to the file.

cloudify_agent/installer/__init__.py
```diff
@@ -353,7 +367,24 @@ def download(self, url, destination=None):
...
+                certificate_file=self._get_certificate_file())
 
     def move(self, source, target):
         self.runner.move(source, target)
+
+    def _get_certificate_file(self):
+        cert_file = self.cloudify_agent['local_rest_cert_file']
+        cert_content = self.cloudify_agent.get('rest_cert_content')
+
+        if not cert_content:
```
dankilman: if the user has an image with the certificate already in place (or configured in userdata) then this call should not return `None`, no?

cloudify_agent/installer/__init__.py
dankilman: this actually resolves https://cloudifysource.atlassian.net/browse/CFY-5779 as well

cloudify_agent/installer/__init__.py
tehasdf: the cert has "\\n" instead of newlines in it - ie. the two characters "backslash, lowercase letter N", instead of a single character "\n" - ie. "newline". `.decode('string-escape')` decodes the escape sequences, ie. turns "\\n" into "\n" (it would also eg. turn "\\r" into "\r", or "\\x20" into " " - but the cert is of course pem-encoded, so the only escape sequence inside it, is "\\n")

cloudify_agent/installer/__init__.py
```diff
@@ -353,7 +377,20 @@ def download(self, url, destination=None):
                 url,
                 output_path=destination,
                 skip_verification=not verify_cert,
-                certificate_file=local_cert_file)
+                certificate_file=self.cloudify_agent.get(
```
dankilman: doesn't seem like this change does much :)

tehasdf: Right. Revered that line.

PR #147, opened 2016-07-29
CFY-5756 Default locations of certs are taken from managers config

xwombat:


cloudify_agent/tests/installer/config/test_configuration.py
```diff
@@ -36,6 +36,10 @@ class TestConfiguration(BaseTest):
     def test_prepare(self):
 
         cloudify_agent = {'local': True}
+
+        os.environ['AGENT_REST_CERT_PATH'] = 'rest/cert/path'
```
dusking: need to move the env vars to "tests/__init__.py", add it to the agent_env_vars dict..

PR #148, opened 2016-08-01
CFY-5135 Upload cert to the agent host before downloading agent package

tehasdf:


tehasdf:
needs #145

cloudify_agent/installer/runners/local_runner.py
```diff
@@ -0,0 +1,45 @@
...
+                 certificate_file=None, **attributes):
+        verify = not skip_verification
+        if certificate_file:
+            verify = certificate_file
+        response = requests.get(url, stream=True, verify=verify)
+        if not response.ok:
+            raise HttpException(url, response.status_code, response.reason)
+
+        if output_path:
+            destination_file = open(output_path, 'w')
```
dankilman: `wb`

dankilman: actually, i've now noticed that 3.4.1-build doesn't have it as well.
we made this exact change somewhere else, where was it?

PR #149, opened 2016-08-01
adding required env vars

noakup:


cloudify_agent/installer/__init__.py
```diff
@@ -198,14 +198,14 @@ def _create_agent_env(self):
...
             env.CLOUDIFY_BROKER_PORT: self.cloudify_agent.get('broker_port'),
-            env.CLOUDIFY_FILE_SERVER_PORT:
-                self.cloudify_agent['file_server_port'],
-            env.CLOUDIFY_FILE_SERVER_PROTOCOL:
-                self.cloudify_agent.get('file_server_protocol'),
             env.CLOUDIFY_REST_PORT:
                 self.cloudify_agent.get('rest_port'),
             env.CLOUDIFY_REST_PROTOCOL: self.cloudify_agent.get(
                 'rest_protocol'),
+            env.CLOUDIFY_FILE_SERVER_PORT:
```
dankilman: why is this changed to get?

cloudify_agent/installer/config/configuration.py
```diff
@@ -164,7 +164,6 @@ def _cfy_agent_attributes_no_defaults(cloudify_agent):
             cloudify_utils.get_manager_file_server_host()
 
     if not cloudify_agent.get('file_server_port'):
-        # by default, file_server_port is set by an environment variable
```
dankilman: why is this comment removed?

system_tests/manager/test_agent_installer.py
```diff
@@ -232,22 +234,41 @@ def _manager_host(self):
...
             'rest_host': manager_host,
+            'file_server_host': manager_host,
+            'broker_ip': manager_host,
             'name': name
         }})
     try:
         current_ctx.set(ctx)
-        os.environ['MANAGER_FILE_SERVER_URL'] = 'http://{0}:53229'.format(
-            manager_host)
+        for var_name, value in env_vars.iteritems():
```
dankilman: this can actually be simplified to `os.environ.update(env)`

cloudify_agent/installer/config/configuration.py
noakup: simple because it's true for every other var as well, and pretty clear from the code as it is

cloudify_agent/installer/__init__.py
noakup: since it's the same (in terms of being mandatory or not) as the rest_port and protocol

PR #151, opened 2016-08-01
CFY-5135 Create the directory containing the cert before uploading it

tehasdf:


cloudify_agent/installer/__init__.py
```diff
@@ -387,6 +391,12 @@ def upload_certificate(self):
         cert_content = self.rest_cert_content
 
         if not self.runner.exists(cert_file):
+            cert_dir = os.path.dirname(cert_file)
+            if not self.runner.exists(cert_dir):
+                self.runner.python(
+                    'import os',
+                    'os.makedirs(os.path.expanduser("{0}"))'.format(cert_dir))
```
tehasdf: use python for multiplatform compat, instead of mkdir or other

cloudify_agent/installer/__init__.py
```diff
@@ -387,9 +391,30 @@ def upload_certificate(self):
...
                 os.unlink(f.name)
+
+    def _create_cert_dir(self, cert_dir):
+        """Create the directory containing the manager certificate.
+
+        For cross-platform compatibility, use python to create the directory.
+        """
+
+        try:
+            self.runner.python(
```
tehasdf: this is very ugly but it was the only way i could think of that didnt make me lose my sanity, re:

  - use mkdir on linux and whatever else (?) on windows
  - needs to create nested dirs (like `mkdir -p`)
  - needs to handle already-existing directories
  - needs to handle the `~`

cloudify_agent/installer/__init__.py
```diff
@@ -350,6 +350,10 @@ def upload_certificate(self):
         cert_content = self.rest_cert_content
 
         if not os.path.exists(cert_file):
+            cert_file = os.path.expanduser(cert_file)
```
dankilman: it is logically impossible for exists to return True and at the same time have expanduser have any affect

cloudify_agent/installer/__init__.py
```diff
@@ -387,9 +391,30 @@ def upload_certificate(self):
         cert_content = self.rest_cert_content
 
         if not self.runner.exists(cert_file):
+            cert_dir = os.path.dirname(cert_file)
```
dankilman: if the agent is a windows agent, will running os.path.dirname on the linux manager on certfile do the right thing?

cloudify_agent/installer/__init__.py
tehasdf: right, let's just expanduser before that

cloudify_agent/installer/__init__.py
tehasdf: i think it might but you're right, let's do this on the agent side

cloudify_agent/installer/__init__.py
```diff
@@ -387,9 +391,29 @@ def upload_certificate(self):
         cert_content = self.rest_cert_content
 
         if not self.runner.exists(cert_file):
+
```
dankilman: less whitespace? :)

cloudify_agent/installer/__init__.py
```diff
@@ -387,9 +391,27 @@ def upload_certificate(self):
...
+        try:
+            self.runner.python(
+                'import os',
+                'os.makedirs(os.path.dirname(os.path.expanduser(\'{0}\')))'
+                .format(cert_file))
+        except:
+            # an error was thrown - if the directory does exist, we assume
+            # it was a "directory already exists" error and continue.
+            if not self.runner.exists(cert_file):
```
dankilman: you should be checking the dirname which again leads to the question of behavior on windows.
i guess you can use python again if there is a problem

tehasdf: whoops that's very true

PR #167, opened 2016-09-16
Make broker IP default to rest host

geokala:
Without defaulting to the rest host IP the broker IP will only be populated when an external broker is used (unless there is something else that is making this work?).


geokala:
Note: Still in the process of testing this.

geokala:
Not needed- the property is set elsewhere.

PR #169, opened 2016-09-28
CFY-5996: Consider manager_ip override

isaac-s:


cloudify_agent/installer/config/configuration.py
```diff
@@ -211,6 +211,11 @@ def reinstallation_attributes(cloudify_agent):
     _add_installation_defaults(cloudify_agent)
 
 
+def _get_manager_file_server_url(cloudify_agent):
+    ip = cloudify_agent.get('manager_ip')
+    return 'http://{}:53229'.format(ip)
```
mxmrlv: Maybe it's better to alter the get_manager_file_server_url() to optionally receive an ip. if the ip is none, it will revert to the current usage. either way, i think it's better to write the function in the following way:
```
'{0}://{1}:{2}'.format(get_manager_file_server_protocol(), ip, get_manager_file_server_port())
```

isaac-s: 1. Changing `get_manager_file_server_url()` will involve changing `cloudify-plugins-commons`. That function is a part of a set of functions that seem to only want to deal with the OS environment. It seems to me that for agent properties calculation, we should not touch `get_manager_file_server_url` at all but build on top of it instead.

2. `get_manager_file_server_protocol` and `get_manager_file_server_port` are only available on `master`, it's not in the `3.4.0.2` tag. So I guess a separate PR will have to be opened.

PR #173, opened 2016-12-04
CFY-6104 Update the agent config with the current cluster state

tehasdf:


cloudify_agent/resources/pm/detach/detach.template
```diff
@@ -22,4 +22,5 @@
 --with-gate-keeper \
 --gate-keeper-bucket-size={{ max_workers }} \
 --with-logging-server \
---logging-server-logdir={{ workdir }}/logs
+--logging-server-logdir={{ workdir }}/logs \
+--heartbeat-interval=5
```
tehasdf: `--heartbeat-interval` is not strictly required (we use the AMQP-level heartbeat as well) but it can be useful to force failover/reconnect in weird edge cases

cloudify_agent/app.py
```diff
@@ -81,17 +83,30 @@ def callback(*args, **kwargs):
...
+
 # This attribute is used as the celery App instance.
 # it is referenced in two ways:
 #   1. Celery command line --app options.
 #   2. cloudify.dispatch.dispatch uses it as the 'task' decorator.
 # For app configuration, see cloudify.broker_config.
-
-
-app = Celery()
+app = HeartbeatCelery()
```
tehasdf: doesn't hurt to have it even in the no-cluster case - we'll just gain earlier detection of unclean disconnects

cloudify_agent/app.py
```diff
@@ -124,3 +137,8 @@ def new_excepthook(exception_type, value, the_traceback):
         current_excepthook(exception_type, value, the_traceback)
 
     sys.excepthook = new_excepthook
+    utils.enable_cluster_failover(app)
```
tehasdf: always enable failover and register the `update` task - if the manager isn't a cluster, the failover strategy is equivalent to round-robin (and anyway, it'll just have one broker address), and the task will never be called.

This also means an agent can be reconfigured at runtime - if it was installed with just one manager, it will still be able to switch to a cluster later.

cloudify_agent/app.py
```diff
@@ -124,3 +177,12 @@ def new_excepthook(exception_type, value, the_traceback):
         current_excepthook(exception_type, value, the_traceback)
 
     sys.excepthook = new_excepthook
+
+    os.environ['CLUSTER_SETTINGS_FILE'] = cluster.DEFAULT_CLUSTER_CONFIG_FILE
+    kombu.connection.failover_strategies['check_cluster_config'] = \
+        _make_failover_strategy(daemon_name)
+    app.conf['BROKER_FAILOVER_STRATEGY'] = 'check_cluster_config'
+
```
tehasdf: create the cluster status file on startup as well (if cluster configured)

cloudify_agent/installer/operations.py
```diff
@@ -44,7 +44,6 @@ def prepare_local_installer(cloudify_agent, logger=None):
 
 
 def init_agent_installer(func=None, validate_connection=True):
-
```
tehasdf: fix

cloudify_agent/app.py
```diff
@@ -19,21 +19,28 @@
...
 from cloudify.celery import logging_server
 
 from cloudify_agent.api import utils
+from cloudify_agent.api.factory import DaemonFactory
 
+
+logger = get_logger(__name__)
 LOGFILE_SIZE_BYTES = 5 * 1024 * 1024
 LOGFILE_BACKUP_COUNT = 5
+CLUSTER_CONFIG_FILE = os.path.expanduser('~/.cloudify-cluster')
```
tehasdf: remove

dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
 https://github.com/cloudify-cosmo/cloudify-rest-client/archive/master.zip
-https://github.com/cloudify-cosmo/cloudify-plugins-common/archive/master.zip
+https://github.com/cloudify-cosmo/cloudify-plugins-common/archive/ha-cluster-restclient.zip
```
tehasdf: remove

PR #176, opened 2016-12-08
CFY 5996 Add support for multi-network managers in cloudify agents

mcouthon:


cloudify_agent/shell/commands/daemons.py
```diff
@@ -386,3 +388,39 @@ def _parse_custom_options(options):
...
+    manager_ips = manager_ips.split(',')
+    if not manager_ips:
+        raise exceptions.DaemonMissingMandatoryPropertyError(
+            'Passed empty list of manager IPs'
+        )
+    for host in manager_ips:
+        sock = socket.socket()
+        try:
+            # Try to connect to the host, and return it if successful
+            sock.connect((host, broker_port))
```
dusking: don't forget to close your socket..

tehasdf: it's closed in the `finally` block

btw can this connect call result in a potentially very long wait? maybe we should have a timeout as well?

dusking: in case of success, you won't get there (you have return..)
indeed connect can take a while, but it doesn't say that it's broken,
probably timeout should be configurable..

tehasdf: `finally` is run in both the success case, and the error case :)

dusking: Oops :)

cloudify_agent/shell/commands/manager_ip.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        # The manager ips var is a comma delimited string of IPs (as it passes
+        # through env variables) representing a list - so we split
+        manager_ips = attributes.pop('manager_ips')
+        manager_ips = manager_ips.split(',')
+        manager_ips = ManagerIP._get_sorted_manager_ips(
+            manager_ips,
+            attributes['ip']
+        )
+        if not manager_ips:
```
dusking: better to check after pop

cloudify_agent/shell/commands/daemons.py
```diff
@@ -161,6 +157,11 @@ def create(**params):
     custom_arg = attributes.pop('custom_options', ())
     attributes.update(_parse_custom_options(custom_arg))
     click.echo('Creating...')
+    manager_ip = get_manager_ip(attributes)
```
dusking: will be nicer if it will be more specific, receive manager_ips and agent_ip as arguments, and not obscure "attributes" argument

mcouthon: Well, I need 4 different attributes from there, and I also set some of the attributes back into the dict, so it's easiest for me to just have the whole dict there

PR #196, opened 2017-02-01
CFY-6365 Upload certificate to agent during installation

mcouthon:


cloudify_agent/installer/runners/fabric_runner.py
```diff
@@ -236,7 +236,10 @@ def put_file(self, src, dst=None, sudo=False, **attributes):
         :return: the destination path
         """
 
-        if not dst:
+        if dst:
+            # Make sure the destination folder exists
```
dusking: consider function "verify_folder_in_remote..."

PR #203, opened 2017-02-20
CFY-6380. Install agent under program files

jcollado:
In this PR, the default path to install the agent on a windows machine is set to `C:\Program Files\Cloudify Agents`. The other changes are fixes to make sure that paths are properly quoted when a directory name with an space is used.


cloudify_agent/installer/config/configuration.py
```diff
@@ -307,8 +307,7 @@ def installation_attributes(cloudify_agent, runner):
                 # 'pwd' module does not exists in a windows python
                 # installation.
                 # TODO - maybe use some environment variables heuristics?
```
tehasdf: if we're not installing into the home dir anymore, this comment is obsolete, i suppose remove it

jcollado: Thanks for pointing that out. Updated comment in: 3aa86b4

idanmo:
Once we get this verified in 4.0, please create a similar PR for 3.4.2-build branch.

PR #209, opened 2017-02-22
CFY 6381. Update storage directory default

jcollado:
In this PR, the [`apddirs`](https://pypi.python.org/pypi/appdirs/1.4.0) library is used to get a directory to store cloudify agent data across different platforms.

In the test with windows, instead of using `C:\Users\Admin\.cfy-agent`, it now uses `C:\Users\cloudbase-init\AppData\Local\Cloudify\cloudify-agent`

### DIFF
--- a/cloudify_agent/api/utils.py
+++ b/cloudify_agent/api/utils.py
@@ -98,7 +98,7 @@ class _Internal(object):
             return cls.get_daemon_storage_dir()
         if username is None and cls.CLOUDIFY_DAEMON_USER_KEY in os.environ:
             username = cls.get_daemon_user()
-        return appdirs.user_data_dir('cloudify-agent', 'Cloudify')
+        return appdirs.site_data_dir('cloudify-agent', 'Cloudify')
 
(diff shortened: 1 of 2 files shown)


jcollado:
@isaac-s No problem, I see that if I change `user_data_dir` with `site_data_dir` what I get is:
`C:\ProgramData\Cloudify\cloudify-agent\`

Fixed in a6a77a7

isaac-s:
I'll take that as a "no"

jcollado:
@isaac-s Sorry, I didn't see your comment. If what you mean is using a system-wide directory in linux as well, I tried that, but the installation failed because it needed root permissions to create the directory. That would have required executing the installation as root and which in turn could have caused permission problems later when trying to write to a directory owned by root. Nothing that cannot be fixed, but we have decided to go this way for now to make sure this small change doesn't cause any regression.

isaac-s:
OK, I see. Thank you! (Still, I think we should address this going forward)

PR #214, opened 2017-03-01
CFY-6492 Make the file server always secured

mcouthon:
* The file server is now always internally accessed via the REST's port (with a /resources suffix), and is then redirected to  its internal port
 * All the folder names now reside in cloudify-plugins-common.constants (seeing as how they are indeed constants)
 * All the workers (agent + mgmt) now need to have much fewer params regarding the file server


cloudify_agent/tests/resources/blueprints/agent-from-package/scripts/create-package.py
```diff
@@ -25,7 +25,10 @@
 file_server_port = ctx.node.properties['file_server_port']
 
 package_name = create_agent_package(resource_base, config, ctx.logger)
-package_url = 'http://localhost:{0}/{1}'.format(file_server_port, package_name)
+package_url = 'http://localhost:{0}/{1}'.format(
```
dusking: should it be https or not?

mcouthon: Checked it - this is a local file server, spun especially (and only) for the test. It is passed via the blueprint inputs to this script. So this is not a big.

PR #223, opened 2017-03-08
CFY-6526 Allow different certs for every node in the cluster

tehasdf:


cloudify_agent/app.py
```diff
@@ -116,10 +117,25 @@ def _strategy(initial_brokers):
                 nodes = cluster.get_cluster_nodes()
                 for node in nodes:
                     _set_master(daemon_name, node)
-                    broker_url = 'amqp://{0}:{1}@{2}:5672//'.format(
+
+                    ssl_enabled = node.get('broker_ssl_enabled', False)
```
dusking: Always true, at least by default true..

cloudify_agent/api/utils.py
```diff
@@ -226,6 +225,7 @@ def get_celery_client(broker_url,
             'ca_certs': broker_ssl_cert_path,
             'cert_reqs': ssl.CERT_REQUIRED,
         }
+    celery_client.pool.connection.ensure_connection()
```
tehasdf: this needs to be AFTER ssl settings are set up

cloudify_agent/app.py
tehasdf: lets make this true after we generate rabbitmq certs

cloudify_agent/app.py
```diff
@@ -99,10 +99,14 @@ def callback(*args, **kwargs):
 
 def _set_master(daemon_name, node):
     factory = DaemonFactory()
-    daemon = factory.load(daemon_name)
+    try:
+        daemon = factory.load(daemon_name)
+    except exceptions.DaemonNotFoundError:
+        return
```
tehasdf: this mostly only happens in tests

PR #224, opened 2017-03-08
Failover strategy: ignore empty broker urls

tehasdf:


cloudify_agent/app.py
```diff
@@ -114,6 +114,7 @@ def _set_master(daemon_name, node):
 def _make_failover_strategy(daemon_name):
     def _strategy(initial_brokers):
         logger.debug('Failover strategy: searching for a new rabbitmq server')
+        initial_brokers = [broker for broker in initial_brokers if broker]
```
tehasdf: There must be more than one initial broker for celery to ever use the failover mechanism, so we add one empty broker url; here, we need to filter that empty url out.
See also https://github.com/cloudify-cosmo/cloudify-plugins-common/blob/master/cloudify/broker_config.py#L84-L90

PR #234, opened 2017-03-14
CFY 6380. Add missing quotes for winrm scenario

jcollado:
In this PR, a couple of quotes related to the installation of the cloudify agent on windows using winrm have been added to support paths with spaces.

### DIFF
--- a/cloudify_agent/installer/__init__.py
+++ b/cloudify_agent/installer/__init__.py
@@ -49,7 +49,7 @@ class AgentInstaller(object):
         if execution_env is None:
             execution_env = {}
 
-        # TBD: Investigate why adding quotes don't work on linux
+        # TBD: Investigate why adding quotes doesn't work on linux
         if self.cloudify_agent.get('windows', False):
             full_command = '"{0}" {1}'
         else:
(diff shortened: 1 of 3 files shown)


cloudify_agent/installer/__init__.py
```diff
@@ -49,7 +49,7 @@ def run_agent_command(self, command, execution_env=None):
         if execution_env is None:
             execution_env = {}
 
-        # TBD: Investigate why adding quotes don't work on linux
+        # TBD: Investigate why adding quotes doesn't work on linux
```
tehasdf: Quotes don't work on linux because our `LinuxInstallerMixin` command is `/venv/bin/python /venv/bin/cfy-agent` (ie. python is the executable, and cfy-agent is the first argument) - and a shell doesn't allow commands that are quoted like `"executable arg" arg2`


Note: if someone was able to prove the `/venv/bin/python` prefix isn't necessary, and `/venv/bin/cfy-agent` is enough (ie. `cfy-agent` is executable in all cases - source or package install), then they would be able to remove the `/../bin/python` from `LinuxInstallerMixin`, and they could add quotes for linux as well.

jcollado: Thanks for your comment. Looking at the code, maybe the problem could be fixed by removing the quotes in `AgentInstaller` and moving them to `Windows/LinuxInstallerMixing.cfy_agent_path`. This way, it should be possible to quote both `python` and `cfy-agent`. I'll give it a try in a separate PR.

tehasdf: good idea!

PR #240, opened 2017-03-22
CFY-6619 Use a per-daemon lock when storing cluster settings

tehasdf:


cloudify_agent/api/factory.py
```diff
@@ -197,7 +197,7 @@ def save(self, daemon):
                 daemon.name)
         )
 
-        if getattr(daemon, 'cluster'):
+        if getattr(daemon, 'cluster', None):
```
tehasdf: 2-arg getattr would throw anyway if the attr wasn't there, lol :p

cloudify_agent/app.py
```diff
@@ -19,15 +19,18 @@
...
 import traceback
+import tempfile
 import logging
 import logging.handlers
 import kombu.connection
 
 from celery import Celery, signals
 from celery.utils.log import ColorFormatter, get_logger
 from celery.worker.loops import asynloop
+from fasteners import InterProcessLock
```
tehasdf: fasteners is already installed in agent

cloudify_agent/app.py
```diff
@@ -107,8 +117,9 @@ def _set_master(daemon_name, node):
     daemon.broker_user = node['broker_user']
     daemon.broker_pass = node['broker_pass']
     daemon.broker_ssl_cert_path = node.get('internal_cert_path')
-    factory.save(daemon)
-    cluster.set_cluster_active(node)
+    with _cluster_settings_lock(daemon_name):
```
tehasdf: locking every write to the settings file here

cloudify_agent/api/factory.py
```diff
@@ -196,18 +196,18 @@ def save(self, daemon):
             self.storage, '{0}.json'.format(
                 daemon.name)
         )
+        cluster_path = os.path.join(
+            self.storage, 'cluster-{0}.json'.format(
+                daemon.name)
+        )
+        daemon.cluster_settings_path = cluster_path
```
tehasdf: set that path always - even if we don't have a cluster yet, we can join it later without restarting the agent - and by then, we'll need to have already decided on the path

PR #244, opened 2017-03-27
CFY-6635 Add client certificate validation to userdata scripts

mcouthon:


cloudify_agent/resources/script/windows.ps1.template
```diff
@@ -4,28 +4,35 @@ $ErrorActionPreference = "Stop"
 
 function Download($Url, $OutputPath)
 {
-    New-Item -ItemType directory -Path (Split-Path $OutputPath)
+    # Disable server side certificate validation (client side validations still exist)
+    [System.Net.ServicePointManager]::ServerCertificateValidationCallback = {$true}
```
idanmo: Why is this needed?
AFAIK it means "trust all".

cloudify_agent/resources/script/windows.ps1.template
```diff
@@ -4,28 +4,35 @@ $ErrorActionPreference = "Stop"
...
+
+    $WebRequest.Headers.Add('{{ auth_token_header }}', '{{ auth_token_value }}')
+
+    # Create a stream from the HTTP response and create the output file
+    $ResponseStream = $WebRequest.GetResponse().GetResponseStream()
+    $OutputFile = [System.IO.File]::Create($OutputPath)
+    $Buffer = New-Object Byte[] 1024
+
+    # Write the stream to the output file in 1KB chunks
+    While (($BytesRead = $ResponseStream.Read($Buffer, 0, $Buffer.Length)) -gt 0) {
```
idanmo: Perhaps extract the download functionality to a function..

cloudify_agent/resources/script/windows.ps1.template
```diff
@@ -4,28 +4,35 @@ $ErrorActionPreference = "Stop"
...
+        $OutputFile.Write($Buffer, 0, $BytesRead)
+    }
+    $OutputFile.Close()
```
idanmo: It probably doesn't matter much, but you can use the Try..Finally pattern here and call $OutputFile.Close() in the finally block.

PR #249, opened 2017-05-09
CFY-6843 Change initd service name per daemon

tehasdf:
Instead of hardcoding the service name "celeryd", let's use
a different service name for each agent. This allows installing
more than one agent on systems that check that name
(eg. ubuntu 16.04)


geokala:
Closed: As this PR is more than six months old without having been merged. Please reopen (and rebase if necessary) if it is still relevant.

PR #253, opened 2017-06-01
CFY-6965: tempdir override for executables

isaac-s:


isaac-s:
The agent installer uses `/tmp` to write the temporary `disable-requiretty.sh` into it & then execute. This PR changes it so it looks for the executable tempdir first.

PR #264, opened 2017-06-22
CFY-7016 When installing source plugins, make the dir group-writable

tehasdf:


tehasdf:
obsoleted by cloudify-cosmo/cloudify-premium#208

PR #265, opened 2017-06-22
CFY-6966 Fix recreate deployment envs

geokala:


geokala:
Depends on:
https://github.com/cloudify-cosmo/cloudify-manager/pull/898
https://github.com/cloudify-cosmo/cloudify-plugins-common/pull/338

cloudify_agent/operations.py
```diff
@@ -48,6 +49,11 @@
 
 @operation
 def install_plugins(plugins, **_):
+    tenant_mask = False
+    if ctx.tenant_name != workflow_ctx.tenant_name:
```
idanmo: Can you explain this scenario?
Isn't it possible to propagate the tenant information?
ctx._context is internal, its modification should be somehow encapsulated...

geokala: The problem is that any updates to execution status (e.g. if the workflow fails) will still require this same logic. See comments on https://github.com/cloudify-cosmo/cloudify-plugins-common/pull/338 for why I think we probably don't want to bed this in much further than it is.

PR #273, opened 2017-07-19
CFY-7086 Send agent config as dict instead of object

mcouthon:
Otherwise, the code on the other side won't be able to deal with it


mcouthon:
I don't want to wait for the tests again :P
I'll do it in my other agent branch

PR #278, opened 2017-07-26
CFY-7110 Add an explicit --cluster to cfy-agent

tehasdf:


tehasdf:
note to self, remember to remove the dev-reqs commit when merging

PR #289, opened 2017-08-07
CFY-7142. Fix command line too long

jcollado:
In this PR, the code that sends the agent installation script to a windows server is updated to split the script into chunks.

The reason to split the script is that to the script is included as part of a powershell command, so its size is restricted by the maximum command line length supported in windows to avoid getting a command line too long error.

### DIFF
--- a/cloudify_agent/installer/runners/winrm_runner.py
+++ b/cloudify_agent/installer/runners/winrm_runner.py
@@ -398,7 +398,7 @@ $webClient.Downloadfile('{2}', '{3}')""".format(
         :rtype WinRMCommandExecutionResponse.
         """
         # Escape single quotes, since the contents is surrounded by them
-        contents = contents.replace("'", "`'")
+        contents = contents.replace("'", "''")
 
         # Split content into chunks to avoid command line too long error
         # maximum allowed commmand line length should be 2047 in old windows:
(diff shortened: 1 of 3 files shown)


cloudify_agent/installer/runners/winrm_runner.py
```diff
@@ -470,12 +483,50 @@ def close(self):
...
+
+    :param contents:
+        The contents of a file that exceeds the maximum command line length in
+        windows.
+    :type content: str
+    :returns: The same content in chunks that won't exceed the limit
+    :rtype: list[str]
+
+    """
+    # Split content into chunks to avoid command line too long error
```
mcouthon: This seems to be repeating the text in the docstring

cloudify_agent/installer/runners/winrm_runner.py
```diff
@@ -470,12 +483,50 @@ def close(self):
...
+    # maximum allowed commmand line length should be 2047 in old windows:
+    # https://support.microsoft.com/en-us/help/830473/command-prompt-cmd--exe-command-line-string-limitation
+    max_size = 2000
+    separator = '\r\n'
+
+    def join_lines(lines, line):
```
mcouthon: This is pretty neat! But I think I'd like a unittest for this particular function - just to see that we get exactly the results we want. And maybe play around with different strings and escape characters inside the split string

cloudify_agent/installer/runners/winrm_runner.py
jcollado: Fixed in ee019f5

cloudify_agent/installer/runners/winrm_runner.py
jcollado: Fixed in 4bb35d6

PR #291, opened 2017-08-10
CFY-7107. Delete agent installation script (winrm)

jcollado:
In this PR, the agent installation script is successfully deleted from the remote windows machine after installation or failure during the script execution.

### DIFF
--- a/cloudify_agent/installer/runners/winrm_runner.py
+++ b/cloudify_agent/installer/runners/winrm_runner.py
@@ -483,6 +483,14 @@ $webClient.Downloadfile('{2}', '{3}')""".format(
         pass
 
     def run_script(self, script_path):
+        """Upload script to remote instances, execute it and delete it.
+
+        :param script_path: Local path to script
+        :type script_path: str
+        :return: Script execution output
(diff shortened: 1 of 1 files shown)


cloudify_agent/installer/runners/winrm_runner.py
```diff
@@ -483,13 +485,24 @@ def close(self):
...
             self.get_temp_dir(),
             ntpath.basename(script_path),
         )
-        self.put_file(script_path, remote_path)
-        result = self.run(remote_path, powershell=True)
-        self.delete(ntpath.dirname(remote_path))
+        try:
+            self.put_file(script_path, remote_path)
+            result = self.run(remote_path, powershell=True)
+            self.logger.critical('Remote debugging!')
```
mcouthon: Probably shouldn't be here :)

jcollado: Yes, absolutely. Thanks for pointing that out.

jcollado: Fixed in f5eb4ff

PR #300, opened 2017-08-17
CFY-7105. Pass install parameter to script builder

jcollado:
In this PR, the `install` parameter is passed all the way from the agent create operation to the method that renders the agent installer script template to let the user decide if it's needed to install the agent in the remote instance or just configure and start an agent configured externally.


jcollado:
This is not the best solution since it requires to pass `install` parameter explicitly along multiple functions. A different implementation in which the parameter is passed through the blueprint will be investigated.

PR #301, opened 2017-08-18
CFY-7156. Remove agent script file after a timeout

jcollado:
In this PR, when an agent script is generated,  a transient systemd timer unit is created to remove that file automatically after a timeout currently hardcoded to 5 minutes (timeout should be made configurable in a future PR).


mcouthon:
Taken care of in another way [here](https://github.com/cloudify-cosmo/cloudify-agent/pull/306)

PR #303, opened 2017-08-21
CFY-7168 Add cert in download script for userdata/agent upgrade

mcouthon:
In the cases where the download script is used, avoid having the
certificate in the install script, instead using the same path to
which it was downloaded in the download script

### DIFF
--- a/cloudify_agent/installer/script.py
+++ b/cloudify_agent/installer/script.py
@@ -51,7 +51,7 @@ class AgentInstallationScriptBuilder(AgentInstaller):
             self.init_script_filename = '{0}.sh'.format(uuid.uuid4())
             self.custom_env_path = '{0}/custom_agent_env.sh'.format(basedir)
 
-    def install_script(self):
+    def install_script(self, add_ssl_cert=True):
         """Render the agent installation script.
         :return: Install script content
         :rtype: str
@@ -62,7 +62,7 @@ class AgentInstallationScriptBuilder(AgentInstaller):
             lstrip_blocks=True,
         )
         # Called before creating the agent env to populate all the variables
-        local_rest_content = self._get_local_cert_content()
+        cert_content = self._get_local_cert_content() if add_ssl_cert else ''
         remote_ssl_cert_path = self._get_remote_ssl_cert_path()
         # Called before rendering the template to populate all the variables
         daemon_env = self._create_agent_env()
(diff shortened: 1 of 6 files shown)


mcouthon:
It should be noted that the goal of this PR is to make the download script more secure by not skipping certificate validation when downloading the install script.

cloudify_agent/resources/script/linux-download.sh.template
```diff
@@ -2,16 +2,21 @@
...
+
 download()
 {
     SCRIPT_NAME=$1
 
-    # Downloading with '--insecure' flags, because at this point we don't have
-    # the SSL certificates
     if command -v wget > /dev/null 2>&1; then
-        wget {{ link }} -O ${SCRIPT_NAME} --no-check-certificate
+        wget {{ link }} -O ${SCRIPT_NAME} --ca-certificate {{ ssl_cert_path }}
```
tehasdf: maybe "quote" the paths in this file? in case they contain spaces

cloudify_agent/tests/__init__.py
```diff
@@ -66,15 +66,24 @@ def get_local_cert_path(temp_folder):
         return f.name
 
     @staticmethod
+    def _clean_cert(cert_content):
+        """ Strip any whitespaces, and normalize the string on windows """
+
+        cert_content = cert_content.strip()
+        cert_content = cert_content.replace('\r\n', '\n').replace('\r', '\n')
```
tehasdf: i think simply opening the file with universal newlines ('U') mode would do this as well?

cloudify_agent/installer/script.py
```diff
@@ -109,39 +110,38 @@ def _get_script_path_and_url(self, script_filename):
...
         1. Path where the install script resides in the file server
         2. URL where the install script can be downloaded
         :rtype: (str, str)
         """
-        return self._script_download_link(is_install_script=False)
+        script_filename = self.install_script_filename
+        script_content = self.install_script(add_ssl_cert=add_ssl_cert)
+        return self._get_script_path_and_url(script_filename, script_content)
+
+    def _get_script_path_and_url(self, script_filename, script_content):
```
tehasdf: kinda having a hard time following the changes in this module, if possible maybe it makes sense to separate commits

mcouthon: Sorry, but it would be  really difficult to separate the changes in this file into more than one coherent commit

cloudify_agent/resources/script/linux-download.sh.template
mcouthon: Should be fine, really - it's never been quoted in other places (like `linux.sh.template`)

PR #304, opened 2017-08-22
CFY-7105 Don't install agent via install script in provided mode

mcouthon:


cloudify_agent/tests/installer/script/test_install_script.py
```diff
@@ -37,28 +37,35 @@ def _serve(self):
 
     def setUp(self):
         super(BaseInstallScriptTest, self).setUp()
-        ctx = MockCloudifyContext(
-            node_id='d',
-            properties={'agent_config': {
+        self._set_mock_context()
+        self.addCleanup(lambda: current_ctx.clear())
```
tehasdf: can just `.addCleanup(current_ctx.clear)`

PR #306, opened 2017-08-25
CFY-7156 Delete init/install scripts after they've been used

tehasdf:


cloudify_agent/installer/script.py
```diff
@@ -203,3 +207,27 @@ def install_script_path(cloudify_agent):
...
         os.remove(script_path)
+
+
+def _cleanup_after_installation(path):
+    """Mark path to be deleted after agent installation.
+
+    This simply adds the path to cloudify_agent inside runtime properties,
+    so that it can be removed later.
+    """
+    cloudify_agent = ctx.instance.runtime_properties.get('cloudify_agent', {})
```
mcouthon: This seems unnecessary - you could just pass `self.cloudify_agent` from the above class

cloudify_agent/installer/script.py
```diff
@@ -203,3 +207,27 @@ def install_script_path(cloudify_agent):
...
+    cleanup = cloudify_agent.get(LOCAL_CLEANUP_PATHS_KEY, [])
+    cleanup.append(path)
+    cloudify_agent[LOCAL_CLEANUP_PATHS_KEY] = cleanup
+    ctx.instance.runtime_properties['cloudify_agent'] = cloudify_agent
+
+
+def cleanup_scripts():
+    """Remove the files that were scheduled for deletion."""
+    cloudify_agent = ctx.instance.runtime_properties.get('cloudify_agent', {})
+    paths = cloudify_agent.pop(LOCAL_CLEANUP_PATHS_KEY, [])
```
mcouthon: Does this pop gets registered by the ctx and the runtime props?

cloudify_agent/installer/script.py
```diff
@@ -203,3 +207,27 @@ def install_script_path(cloudify_agent):
...
+    """Mark path to be deleted after agent installation.
+
+    This simply adds the path to cloudify_agent inside runtime properties,
+    so that it can be removed later.
+    """
+    cloudify_agent = ctx.instance.runtime_properties.get('cloudify_agent', {})
+    cleanup = cloudify_agent.get(LOCAL_CLEANUP_PATHS_KEY, [])
+    cleanup.append(path)
+    cloudify_agent[LOCAL_CLEANUP_PATHS_KEY] = cleanup
+    ctx.instance.runtime_properties['cloudify_agent'] = cloudify_agent
```
mcouthon: Here you can use the `update_agent_runtime_properties` func from `agent_config.py`

cloudify_agent/installer/script.py
tehasdf: it doesn't, added a `update` there

PR #311, opened 2017-09-11
On an error is raised while creating a runner, retry and don't continue

tehasdf:
Before the refactor, we used to run .retry() and not continue trying
to use the runner. Now we need to make sure this behaviour is kept:
we set .retry(), and then we need to make sure no further attempts
to use the runner are done. So, in that case, we simply `return` from
the caller function.

### DIFF
--- a/cloudify_agent/installer/config/agent_config.py
+++ b/cloudify_agent/installer/config/agent_config.py
@@ -54,6 +54,8 @@ def create_agent_config_and_installer(func=None,
             cloudify_agent.set_default_values()
 
         runner = create_runner(cloudify_agent, validate_connection)
+        if not runner:
+            return
         cloudify_agent.set_installation_params(runner)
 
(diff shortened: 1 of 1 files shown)


cloudify_agent/installer/config/agent_config.py
```diff
@@ -54,6 +54,8 @@ def wrapper(*args, **kwargs):
             cloudify_agent.set_default_values()
 
         runner = create_runner(cloudify_agent, validate_connection)
+        if runner is None:
```
mcouthon: `if not runner`? :P

tehasdf:
We need to consider if the retry is even useful at all. Maybe we should remove it altogether. It seems it wasn't critically required when it was added: https://github.com/cloudify-cosmo/cloudify-agent/commit/2179c49c3bd279a90ee935dd2c836b8472ea2bc2

PR #334, opened 2018-01-04
CFY-7759 Allow configuring broker heartbeat

tehasdf:


tehasdf:
This setting is already used in https://github.com/cloudify-cosmo/cloudify-plugins-common/blob/7a0b0390e5c3ff9961dfdf65fecf705ecdb00976/cloudify/broker_config.py#L45

cloudify_agent/api/pm/base.py
```diff
@@ -298,6 +304,8 @@ def create_celery_conf(self):
             'broker_vhost': self.broker_vhost,
             'cluster': self.cluster
         }
+        if self.broker_heartbeat:
+            config['broker_heartbeat'] = self.broker_heartbeat
```
tehasdf: This is the line which adds the setting to the `broker_config.json` file.
The other changes are about getting the value to this place - starting from agent config, via the daemons command.

PR #335, opened 2018-01-08
Fix diamond with clusters

geokala:


geokala:
Tests failing due to problems with cloudify.co SSL cert

PR #337, opened 2018-01-29
CFY-7704 Use TLS 1.2 in the windows download scripts

tehasdf:


geokala:
(note that this will mean that Powershell will use ONLY TLSv1.2 rather than ONLY TLSv1.0)

PR #339, opened 2018-02-07
Use `heartbeat` as --heartbeat-interval

tehasdf:
Using celery heartbeats instead of amqp heartbeats


cloudify_agent/api/pm/nssm.py
```diff
@@ -108,7 +108,8 @@ def create_config(self):
             failure_reset_timeout=self.failure_reset_timeout,
             failure_restart_delay=self.failure_restart_delay,
             cluster_settings_path=self.cluster_settings_path,
-            executable_temp_path=self.executable_temp_path
+            executable_temp_path=self.executable_temp_path,
+            heartbeat=self.broker_heartbeat
```
geokala: Is there any point providing it here? Broker heartbeat doesn't work on windows

PR #340, opened 2018-02-15
CFY-7604: simpler & safer sudoers manipulation

isaac-s:


isaac-s:
Closing — shows user as root when init_script is used

PR #343, opened 2018-02-27
CFY-7939: don't tolerate (most) curl errors. Also, print what you're about to download.

isaac-s:


tehasdf:
contrary to what curl docs say, `-f` is indeed required to have curl exit nonzero on 4xx/5xx codes

PR #355, opened 2018-04-01
CY-215 Pass pip args as a list to wagon.install (#354)

mcouthon:
* CY-215 Pass pip args as a list to wagon.install

* Bump wagon version to 0.6.3


mcouthon:
Merging this though appveyor is failing, to start working on the build

PR #365, opened 2018-04-12
test_operations: use the right fileserver address

tehasdf:
`:53333/resources` is the public address, not `:53229`


geokala:
Closed: As this PR is more than six months old without having been merged. Please reopen (and rebase if necessary) if it is still relevant.

PR #362, opened 2018-04-09
CY-260 modify install script to perform agents transfer

adigrabow:

PR #367, opened 2018-04-16
agent_config.local in agent upgrade as well

tehasdf:


tehasdf:
this is failing CI right now because #362 has not been merged yet, and so the declared `transfer_agent_amqp` operation doesn't exist yet

PR #378, opened 2018-05-07
agents transfer in multi-networks

adigrabow:


cloudify_agent/operations.py
```diff
@@ -557,3 +558,16 @@ def _create_package_url(url, ip):
     after = rest.split(':')[1]
     new = '{before}//{ip}:{after}'.format(before=before, ip=ip, after=after)
     return new
+
+
+def _get_network_ip(manager_ip):
+    network_ip_mapping = yaml.safe_load(manager_ip)
+    # Multi-network mode: a dict containing networks and ips was passed
+    if type(manager_ip) is dict:
```
mcouthon: Use `isinstance` instead

cloudify_agent/operations.py
```diff
@@ -557,3 +558,16 @@ def _create_package_url(url, ip):
...
+        agent = ctx.instance.runtime_properties['cloudify_agent']
+        network_name = agent['network']
+        new_network_ip = network_ip_mapping[network_name]
+        return new_network_ip
+    # Regular mode: the only ip of the new manager was passed
+    else:
```
mcouthon: do `elif` with a `string` (or maybe `basestring`), and `else` that raises an error

cloudify_agent/operations.py
```diff
@@ -480,7 +481,7 @@ def validate_agent_amqp(current_amqp=True, manager_ip=None,
 def transfer_agent_amqp(transfer_agent_timeout=300,
                         manager_ip=None, manager_certificate=None,
                         manager_rest_token=None, **_):
-
+    manager_ip = _get_network_ip(manager_ip)
```
mcouthon: Should probably come after `_validate_agent`

mcouthon: And if you've already validated the agent, and you have the dict, pass it to the function, instead of using runtime props again

cloudify_agent/operations.py
```diff
@@ -557,3 +558,16 @@ def _create_package_url(url, ip):
...
+    network_ip_mapping = yaml.safe_load(manager_ip)
+    # Multi-network mode: a dict containing networks and ips was passed
+    if type(manager_ip) is dict:
+        agent = ctx.instance.runtime_properties['cloudify_agent']
+        network_name = agent['network']
+        new_network_ip = network_ip_mapping[network_name]
+        return new_network_ip
+    # Regular mode: the only ip of the new manager was passed
+    else:
+        return network_ip_mapping
```
mcouthon: It's confusing that a single ip is called `network_ip_mapping`

PR #397, opened 2018-05-16
CY-199 Fix agents upgrade

mcouthon:


cloudify_agent/operations.py
```diff
@@ -168,9 +168,11 @@ def _save_daemon(daemon):
 
 def _set_default_new_agent_config_values(
         old_agent, new_agent, transfer_agent=False):
-    if not transfer_agent:
-        new_agent['name'] = utils.internal.generate_new_agent_name(
-            old_agent['name'])
+    if transfer_agent:
```
tehasdf: tbh maybe it'd be better, instead of having a `transfer_agent` flag to `_set_default_new_agent_config_values` and `create_new_agent_config`, to have a new function (`create_transfer_agent_config`) that would call `create_new_agent_config`, and then overwrite the resulting agent with the transfer-specific changes? (ie. just the name i think)

cloudify_agent/operations.py
```diff
@@ -411,16 +411,17 @@ def create_agent_amqp(install_agent_timeout=300, manager_ip=None,
...
 
     result = _validate_current_amqp()
     if not result['agent_alive']:
         raise RecoverableError('New agent did not start and connect')
 
     # Setting old_cloudify_agent in order to uninstall it later.
     ctx.instance.runtime_properties['old_cloudify_agent'] = agents['old']
-    ctx.instance.runtime_properties['cloudify_agent'] = returned_agent
+    ctx.instance.runtime_properties['cloudify_agent'] = new_agent
+    ctx.instance.update()
```
tehasdf: is there a reason for this? at the very end of the operation, there shouldn't be a reason to call update i dont think?

cloudify_agent/operations.py
```diff
@@ -547,36 +547,19 @@ def transfer_agent_amqp(transfer_agent_timeout=300,
...
-    if transfer_mode:
-        status = ctx.instance.runtime_properties['agent_status']
-        status['agent_alive_crossbroker'] = True
-        ssl_path = agent.get('broker_ssl_cert_path')
-        with open(ssl_path, 'r') as ssl_file:
-            ssl_cert = ssl_file.read()
-        broker_conf['broker_ssl_cert'] = ssl_cert
-    ctx.instance.runtime_properties['cloudify_agent'] = agent
-    ctx.instance.update()
+    return {
```
tehasdf: nice

cloudify_agent/operations.py
```diff
@@ -318,17 +319,33 @@ def _run_script_celery(agent, params, timeout):
...
+        broker_config = agent['broker_config']
+        ssl_cert_path = _get_ssl_cert_path(broker_config)
+        # Using a temp path, so we need to delete it
+        delete_cert_path = True
+    else:
+        broker_config = _get_broker_config(agent)
+        ssl_cert_path = get_local_rest_certificate()
+
+    # TODO: Remove!
+    ctx.logger.error('Broker_config: {0}'.format(broker_config))
```
tehasdf: remember

cloudify_agent/operations.py
```diff
@@ -318,17 +319,33 @@ def _run_script_celery(agent, params, timeout):
...
+    try:
+        yield amqp_client.get_client(
+            amqp_host=broker_config.get('broker_ip'),
```
tehasdf: can we rename things in `_create_broker_config` and do `**` here?

cloudify_agent/tests/test_operations.py
```diff
@@ -110,8 +110,12 @@ def file_server_url(*args, **kwargs):
             return '{0}/resources'.format(http_rest_host({}))
 
         # Need to patch, to avoid broker_ssl_enabled being True
+        @contextmanager
         def get_amqp_client(agent):
-            return get_client()
+            try:
```
tehasdf: no need for the try/finally if it does nothing ;P

cloudify_agent/operations.py
```diff
@@ -530,8 +532,10 @@ def validate_agent_amqp(current_amqp=True, manager_ip=None,
 def transfer_agent_amqp(transfer_agent_timeout=300,
                         manager_ip=None, manager_certificate=None,
                         manager_rest_token=None, **_):
-    _create_broker_config(transfer_mode=True)
```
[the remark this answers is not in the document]
tehasdf: i guess it wasnt needed after all

cloudify_agent/operations.py
mcouthon: the problem with this is that `create_new_agent_config` is called from inside `_run_install_script`, so I'd have to change more than I want to

cloudify_agent/operations.py
mcouthon: fair enough. I get paranoid about this. I'll change it

PR #408, opened 2018-05-30
is_agent_alive: handle ChannelClosed

tehasdf:
That might be thrown when the agent didn't create the exchange yet.
In this case, it's obviously very much NOT alive


tehasdf:
CI red because i cancelled py26 which was with ssh

PR #410, opened 2018-06-04
CY-339 Add systemd daemon tests

tehasdf:


tehasdf:
oh right the issue was that our circle image doesnt have systemd

PR #415, opened 2018-06-12
CY-384 Add a subprocess registry to handlers, for cancelling tasks 

tehasdf:


cloudify_agent/worker.py
```diff
@@ -176,15 +194,73 @@ def new_excepthook(exception_type, value, the_traceback):
...
+    def make_key(self, handler):
+        return handler.ctx.task_id
+
+
+class WorkflowProcessRegistry(_ProcessRegistry):
+    def make_key(self, handler):
+        return handler.ctx.execution_id
+
+    def _stop_process(self, process):
+        time.sleep(10)
```
mcouthon: Let's add a comment here to explain why we're sleeping

cloudify_agent/worker.py
```diff
@@ -176,15 +194,73 @@ def new_excepthook(exception_type, value, the_traceback):
...
+    def unregister(self, handler):
+        self._processes.pop(self.make_key(handler), None)
+
+    def cancel(self, operation_id):
+        p = self._processes.get(operation_id)
+        if p:
+            t = threading.Thread(target=self._stop_process, args=(p, ))
+            t.start()
+
+    def _stop_process(self, process):
```
mcouthon: I guess if you have class A and classes B and C inherit from it, and you only ever use B and C, and C defines method `foo`, then I guess `foo` should only exist in B and not A.
This is a roundabout way to say - let's maybe move `_stop_process` to `OperationProcessRegistry`

cloudify_agent/worker.py
```diff
@@ -176,15 +194,73 @@ def new_excepthook(exception_type, value, the_traceback):
...
+        time.sleep(10)
+        super(WorkflowProcessRegistry, self)._stop_process(process)
+
+
 def make_amqp_worker(args):
     if args.name:
         _setup_excepthook(args.name)
     _setup_logger(args.name)
 
+    operation_registry = OperationProcessRegistry()
```
mcouthon: Is there a reason why those aren't global? Seems a bit weird to create them here and then to pass them to all of the consumers. I think you maybe have already explained this, but I forgot.

cloudify_agent/worker.py
```diff
@@ -160,9 +163,16 @@ def cluster_update_task(self, nodes):
...
 
-    def cancel_workflow_task(self, execution_id):
+    def cancel_workflow_task(self, execution_id, rest_token, tenant):
         logger.info('Cancelling execution {0}'.format(execution_id))
+
+        class CancelCloudifyContext(object):
+            def __init__(self):
+                self.tenant = tenant['name']
+                self.rest_token = rest_token
         self._workflow_registry.cancel(execution_id)
```
mcouthon: Can `cancel` fail? If it does, don't we still want to update the execution status?

cloudify_agent/worker.py
tehasdf: nah C uses it as well (calls super, right)

cloudify_agent/worker.py
tehasdf: i do prefer dependency injection (even manual, without a framework) to globals

cloudify_agent/worker.py
tehasdf: I think it has no real failure modes, no. So if it does fail and the programmer says it shouldn't, it's not safe to continue anyway ;)

PR #428, opened 2018-07-04
CY-430: switch nosetests to pytest runner

tyacbovi:


cloudify_agent/tests/api/plugins/test_installer.py
```diff
@@ -19,6 +19,7 @@
 import platform
 import shutil
 import multiprocessing
+import unittest
```
yarivgiga: better put empty line under it, and order all imports by length

cloudify_agent/tests/api/pm/__init__.py
```diff
@@ -14,13 +14,14 @@
...
 import os
-import nose.tools
+import unittest
 import inspect
 import types
 from functools import wraps
 from mock import _get_target
 from mock import patch
 
+
```
yarivgiga: no need in 2 empty lines.
also, need empty line after line 19.
sort imports by line length.

PR #430, opened 2018-07-10
Use logging level INFO for the ping task timeout

tehasdf:
In case of polling until the agent becomes available, it is normal
for the task to timeout before the agent comes online, no need to
log it on the ERROR level.

### DIFF
--- a/cloudify_agent/api/utils.py
+++ b/cloudify_agent/api/utils.py
@@ -239,9 +239,12 @@ def is_agent_alive(name,
         try:
             response = handler.publish(task, routing_key='service',
                                        timeout=timeout, expiration=expiration)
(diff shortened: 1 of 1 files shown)


cloudify_agent/api/utils.py
```diff
@@ -239,9 +239,12 @@ def is_agent_alive(name,
         try:
             response = handler.publish(task, routing_key='service',
                                        timeout=timeout, expiration=expiration)
-        except (RuntimeError, pika.exceptions.AMQPError) as e:
-            logger.error('Error sending a ping task to {0}: {1}'
-                         .format(name, e))
+        except pika.exceptions.AMQPError as e:
```
isaac-s: Should this really be `info`?
Sounds more like a `warning` to me, no?

PR #442, opened 2018-08-20
CY-540 Small fixes in the hooks consumer

inbala:


cloudify_agent/worker.py
```diff
@@ -290,10 +291,15 @@ def _get_hook(self, event_type):
                         "hook config file doesn't exist".format(event_type))
             return None
         with open(self.HOOKS_CONFIG_PATH) as hooks_conf_file:
-            hooks_conf = yaml.safe_load(hooks_conf_file)['hooks']
+            hooks_conf = yaml.safe_load(hooks_conf_file).get('hooks')
+
+        if not hooks_conf:
```
tehasdf: if we log-and-not-break on this, maybe lets also handle a malformed file too? i mean `try/except ValueError` around the yaml load

PR #447, opened 2018-09-11
Change pip command order

szpotona:


szpotona:
@isaac-s

tehasdf:
we want this in master as well, at least for before we figure out a way to make reqs.txt a first class citizen of source plugins 
#448

isaac-s:
@tehasdf are you taking care of `master`? I don't have time to create a PR for it at the moment.
Also, let's remove all Python-based references to `pip` API and use `subprocess` instead, I believe this will fix the signal-related problems, no?

tehasdf:
@isaac-s I've already merged the #448 I've linked above which does it on master.
Removing imports from pip is something that we should do, because that's blocking us from upgrading pip to above 9.0.1, but it's not in scope for 4.5

isaac-s:
@tehasdf can we have it in scope for 4.6? is there a JIRA?
It's of the first things we need to do.
If I knew about it beforehand I'd have done it for 4.4 long time ago, we must support later `pip` versions.

PR #467, opened 2018-11-07
CY-764 Create and update an agent record in the DB

inbala:


dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip
+https://github.com/cloudify-cosmo/cloudify-common/archive/CY-764-create-agents-table.zip
```
inbala: Change it back before merging

test-requirements.txt
```diff
@@ -2,7 +2,7 @@ testtools
 mock==1.0.1
 unittest2
 FileServer
-https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip
+https://github.com/cloudify-cosmo/cloudify-common/archive/CY-764-create-agents-table.zip
```
inbala: Change it back to master before merging

cloudify_agent/installer/operations.py
```diff
@@ -61,13 +64,22 @@ def create(cloudify_agent, installer, **_):
...
 @create_agent_config_and_installer
 def configure(cloudify_agent, installer, **_):
-    ctx.logger.info('Configuring Agent {0}'.format(cloudify_agent['name']))
-    installer.configure_agent()
+    agent_name = cloudify_agent['name']
+    ctx.logger.info('Configuring Agent {0}'.format(agent_name))
+    utils.update_agent_record(agent_name, AgentState.CONFIGURING)
+    try:
+        installer.configure_agent()
+    except CommandExecutionError, e:
```
tehasdf: don't use the old/deprec syntax (stops working in py3), use `except CommandExecutionError as e` instead

cloudify_agent/installer/operations.py
```diff
@@ -86,10 +100,12 @@ def start(cloudify_agent, **_):
     agent_alive = utils.is_agent_alive(cloudify_agent['queue'], client)
 
     if not agent_alive:
+        utils.update_agent_record(agent_name, AgentState.NONRESPONSIVE)
```
tehasdf: waiting to start is a normal stage during the installation, i'm not sure if we need to set nonresponsive right off the bat

cloudify_agent/installer/operations.py
```diff
@@ -123,5 +145,13 @@ def delete(cloudify_agent, installer, **_):
...
     # no need to handling remote_execution False because this operation is
     # not invoked in that case
-    ctx.logger.info('Restarting Agent {0}'.format(cloudify_agent['name']))
-    installer.restart_agent()
+    agent_name = cloudify_agent['name']
+    ctx.logger.info('Restarting Agent {0}'.format(agent_name))
+    utils.update_agent_record(agent_name, AgentState.RESTARTING)
+    try:
+        installer.restart_agent()
+    except CommandExecutionError, e:
```
tehasdf: use new syntax please

cloudify_agent/tests/test_operations.py
```diff
@@ -91,9 +81,10 @@ def _manager_env(self):
             finally:
                 fs.stop()
 
-    @patch('cloudify.manager.get_rest_client', _MockRestclient)
+    @patch('cloudify_agent.api.utils.common_get_rest_client',
+           return_value=MockRestclient())
     @only_ci
-    def test_install_new_agent(self):
+    def test_install_new_agent(self, _):
```
tehasdf: might as well `*_`

cloudify_agent/operations.py
```diff
@@ -427,6 +428,7 @@ def _stop_old_agent(new_agent, old_agent, timeout):
 
     _run_script(new_agent, script_url, timeout)
     ctx.logger.info('Old Cloudify agent stopped')
+    utils.update_agent_record(old_agent.get('name'), AgentState.STOPPED)
```
tehasdf: there's likely no reason to use `.get` in all those calls - if the name isn't there, then the update will fail anyway. So might as well either fail faster with a keyerror, or have a `if not name: raise ...` inside of `update_agent_record`

inbala: ok so i'll change to `old_agent['name']`

cloudify_agent/installer/operations.py
inbala: I agree that it can happen in the normal stage and then it's misleading.
So I want to change it, but not sure how would I know if it is really nonresponsive?

PR #469, opened 2018-11-14
Re-wrote extract_package_to_dir to use subprocess to make multi-threa…

adigrabow:
…ded snapshot restore work

### DIFF
--- a/cloudify_agent/api/plugins/installer.py
+++ b/cloudify_agent/api/plugins/installer.py
@@ -351,8 +351,8 @@ class PluginInstaller(object):
 
 def extract_package_to_dir(package_url):
     """
-    Using a subprocess to extracts a pip package to a temporary directory.
-    :param: package_url: the URL to the package source.
+    Using a subprocess to extract a pip package to a temporary directory.
+    :param: package_url: the URL of the package source.
     :return: the directory the package was extracted to.
 
     """
@@ -361,6 +361,10 @@ def extract_package_to_dir(package_url):
     runner = LocalCommandRunner()
 
     try:
+        # We run `pip download` command in a subprocess to support
+        # multi-threaded scenario (i.e snapshot restore).
+        # We don't use `curl` because pip can handle different kinds of files,
+        # including .git.
         command = [get_pip_path(), 'download', '-d',
                    archive_dir, '--no-deps', package_url]
         runner.run(command=command)
@@ -390,13 +394,9 @@ def _get_plugin_path(plugin_dir_parent, package_url):
     if len(contents) < 1:
         _remove_tempdir_and_raise_proper_exception(package_url,
                                                    plugin_dir_parent)
-    plugin_dir_name = contents[0][1][0]
+    parent_dir_content = contents[0]
+    plugin_dir_name = parent_dir_content[1][0]
     return os.path.join(plugin_dir_parent, plugin_dir_name)
-
-
-def _assert_list_len_greater_than_value(l, value, package_url, archive_dir):
-    if len(l) < value:
-        _remove_tempdir_and_raise_proper_exception(package_url, archive_dir)
 
 
 def _assert_list_len(l, expected_len, package_url, archive_dir):
@@ -424,8 +424,7 @@ def _get_archive(archive_dir, package_url):
     _assert_list_len(contents, 1, package_url, archive_dir)
     files = contents[0][2]
     _assert_list_len(files, 1, package_url, archive_dir)
-    path = os.path.join(archive_dir, files[0])
-    return path
+    return os.path.join(archive_dir, files[0])
 
 
 def extract_package_name(package_dir):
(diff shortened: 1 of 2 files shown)


cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
 
 def extract_package_to_dir(package_url):
     """
-    Extracts a pip package to a temporary directory.
-
-    :param package_url: the URL to the package source.
-
+    Using a subprocess to extracts a pip package to a temporary directory.
```
mcouthon: s/extracts/extract/

cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
...
+    :param: package_url: the URL to the package source.
```
mcouthon: s/to/of/

cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
...
-        plugin_dir = tempfile.mkdtemp()
-        _patch_pip_download()
-        # Import here, after patch
-        import pip
-        pip.download.unpack_url(link=pip.index.Link(package_url),
-                                location=plugin_dir,
-                                download_dir=None,
-                                only_download=False)
-    except Exception as e:
+        command = [get_pip_path(), 'download', '-d',
```
mcouthon: Let's add some more context here - why are we doing it this way, why we can't just use `curl`, etc.

cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
...
+    """
+    plugin_dir_parent is a directory containing the plugin dir. Since we
+    need the plugin`s dir, we find it's name and concatenate it to the
+    plugin_dir_parent.
+    """
+    contents = list(walk(plugin_dir_parent))
+    if len(contents) < 1:
+        _remove_tempdir_and_raise_proper_exception(package_url,
+                                                   plugin_dir_parent)
+    plugin_dir_name = contents[0][1][0]
```
mcouthon: It's succinct but not very readable. Let's maybe do it in two lines but make it more easy to understand.

cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
...
 
 
+def _get_plugin_path(plugin_dir_parent, package_url):
+    """
+    plugin_dir_parent is a directory containing the plugin dir. Since we
+    need the plugin`s dir, we find it's name and concatenate it to the
+    plugin_dir_parent.
+    """
+    contents = list(walk(plugin_dir_parent))
+    if len(contents) < 1:
```
mcouthon: What if `len(contents) > 1`?

cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
...
+        _remove_tempdir_and_raise_proper_exception(package_url,
+                                                   plugin_dir_parent)
+    plugin_dir_name = contents[0][1][0]
+    return os.path.join(plugin_dir_parent, plugin_dir_name)
+
+
+def _assert_list_len_greater_than_value(l, value, package_url, archive_dir):
```
mcouthon: Doesn't seem to be used anywhere

cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
...
+    """
+    archive_dir contains a zip file with the plugin directory. This function
+    finds the name of that zip file and returns the full path to it (the full
+    path is required in order to extract it)
+    """
+    contents = list(walk(archive_dir))
+    _assert_list_len(contents, 1, package_url, archive_dir)
+    files = contents[0][2]
+    _assert_list_len(files, 1, package_url, archive_dir)
+    path = os.path.join(archive_dir, files[0])
```
mcouthon: Just do `return os.path.join(archive_dir, files[0])`

cloudify_agent/api/plugins/installer.py
mcouthon: Why not use `_assert_list_len` here? It seems applicable

cloudify_agent/api/plugins/installer.py
```diff
@@ -350,81 +351,83 @@ def _rmtree(path):
...
+        'the documentation.'.format(package_url))
+
+
+def _get_archive(archive_dir, package_url):
+    """
+    archive_dir contains a zip file with the plugin directory. This function
+    finds the name of that zip file and returns the full path to it (the full
+    path is required in order to extract it)
+    """
+    contents = list(walk(archive_dir))
```
mcouthon: We're doing it twice. Let's have a `_get_dir_contents` function that does the walk, validates that the list has 1 element, and returns it, and use it both here and in `_get_plugin_path`.

PR #474, opened 2018-12-04
CY-780 Store the heartbeat interval in broker_config.json

tehasdf:


cloudify_agent/api/pm/base.py
```diff
@@ -276,7 +276,8 @@ def create_broker_conf(self):
             'broker_password': self.broker_pass,
             'broker_hostname': self.broker_ip,
             'broker_vhost': self.broker_vhost,
-            'cluster': self.cluster
+            'cluster': self.cluster,
+            'broker_heartbeat': self.heartbeat
```
geokala: If you add a trailing comma here then the next time we need to add something to this it'll just be a +1

PR #480, opened 2018-12-26
Don't delete rabbitmq user of proxy agent because it doesn't exist

inbala:


cloudify_agent/installer/operations.py
```diff
@@ -139,7 +139,7 @@ def delete(cloudify_agent, installer, **_):
     update_agent_record(cloudify_agent, AgentState.DELETED)
 
     # TODO: Delete the RabbitMQ queue after deleting the agent
-    delete_agent_rabbitmq_user(cloudify_agent['name'])
+    delete_agent_rabbitmq_user(cloudify_agent)
```
tehasdf: Hmm this has to do with proxy agents, how? :P

tehasdf: Ahh I see due to cloudify-cosmo/cloudify-common#178 , cool.

inbala: Yes, thanks :)

PR #482, opened 2019-01-15
Check if scheduled exec can start, if not - queue it

adigrabow:


cloudify_agent/worker.py
```diff
@@ -139,6 +137,27 @@ def handle_task(self, full_task):
         self._print_task(ctx, 'Finished handling', status)
         return result
 
+    @staticmethod
+    def is_scheduled_execution(full_task):
+        """
+        If a task contains a `dead-letter-exchange` (dlx_id) information it
+        means it was scheduled
+        """
+        if full_task.get('dlx_id'):
```
tyacbovi: can be one liner

PR #486, opened 2019-01-30
CY-1013 Use agent's tmpdir for the installation scripts

tehasdf:


cloudify_agent/installer/config/agent_config.py
```diff
@@ -123,6 +123,13 @@ def has_installer(self):
     def is_windows(self):
         return self['windows']
 
+    @property
+    def tmpdir(self):
+        try:
+            return self['env'][cloudify_utils.CFY_EXEC_TEMPDIR_ENVVAR]
```
isaac-s: Why not use `get_exec_tempdir`?

tehasdf: @isaac-s because that would return the tmpdir of the mgmtworker, and here, we want whatever is declared in the blueprint (`agent_config: extra: env: CFY_EXEC_TMP: ..`; subject to change if we make it a more-official-non-extra param)

PR #488, opened 2019-02-06
Added support for service_user and service_password on Windows

isaac-s:


cloudify_agent/api/pm/base.py
```diff
@@ -249,8 +249,7 @@ def __init__(self, logger=None, **params):
                                         defaults.LOG_FILE_SIZE)
         self.log_max_history = params.get('log_max_history',
                                           defaults.LOG_BACKUPS)
-        self.executable_temp_path = params.get('executable_temp_path') or \
-            get_exec_tempdir()
+        self.executable_temp_path = params.get('executable_temp_path')
```
isaac-s: This is needed because of a bug that's been hidden so far. If the installation method is `init_script`, then this code on Windows is always executed as `Administrator`. If `params.get('executable_temp_path')` returns `None`, then the template will end up receiving `Administrator`'s tempdir (under `C:\Users\Administrator\AppData\...`).

This means that, when this agent creates temporary files to run scripts, it will try to create them in `Administrator`'s tempdir. It works fine when `LOCALSYSTEM` is running the service, but fails when the user running the service has no permissions to access `Administrator`'s tempdir.

The correct way to approach this is to not render `CFY_EXEC_TEMP` at all, if it's not explicitly defined.

This fix is complemented with a fix (see below) to `nssm.conf.template`, surrounding the rendering of `CFY_EXEC_TEMP` with a Jinja `if` statement.

cloudify_agent/installer/__init__.py
```diff
@@ -187,7 +194,7 @@ def _create_process_management_options(self):
         # environment variable
         process_management.pop('name')
         for key, value in process_management.iteritems():
-            options.append('--{0}={1}'.format(key, value))
+            options.append("--{0}={1}".format(key, quote(value)))
```
isaac-s: This is required because the service password may contain something like a dollar sign. Since the `nssm.conf` file is actually rendered on the manager side, this results in a command-line such as:

`cfy-agent daemons configure ... ... ... --service_password=password$dollar`

Which results in attempting evaluation of `$dollar` as an environment variable.

Long story short... the parameter value must be commandline-quoted.

cloudify_agent/resources/pm/nssm/nssm.conf.template
```diff
@@ -26,11 +26,18 @@ AGENT_LOG_LEVEL="{{ log_level }}" ^
 AGENT_WORK_DIR="{{ workdir }}" ^ {{ custom_environment }} ^
 AGENT_LOG_MAX_BYTES="{{ log_max_bytes }}" ^
 AGENT_LOG_MAX_HISTORY="{{ log_max_history }}" ^
+{% if executable_temp_path %}
```
isaac-s: See review comment above.

cloudify_agent/api/pm/base.py
geokala: So if this is set to None, where do temp files end up when the agent creates them?

cloudify_agent/installer/__init__.py
geokala: Is it worth raising a separate Jira to make this file be generated as text and dumped into a file rather than rendered via a terminal? (not part of this issue, just a query)

cloudify_agent/api/pm/base.py
geokala: Oh, I just saw the comment below, ignore this.

cloudify_agent/installer/__init__.py
isaac-s: @geokala not sure I understand....

PR #498, opened 2019-02-14
CY-1051: Properly handle Windows agent installation problems

isaac-s:
Including fixing logging as well (see my comments in the code itself below).

### DIFF
--- a/cloudify_agent/installer/operations.py
+++ b/cloudify_agent/installer/operations.py
@@ -17,7 +17,8 @@ from cloudify.decorators import operation
 from cloudify.amqp_client import get_client
 from cloudify.models_states import AgentState
 from cloudify import ctx, utils as cloudify_utils
-from cloudify.exceptions import CommandExecutionError
+from cloudify.exceptions import (CommandExecutionError,
+                                 CommandExecutionException)
 from cloudify.agent_utils import (create_agent_record,
                                   update_agent_record,
                                   delete_agent_rabbitmq_user)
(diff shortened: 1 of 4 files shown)


cloudify_agent/installer/operations.py
```diff
@@ -40,13 +41,10 @@ def create(cloudify_agent, installer, **_):
             ctx.logger.info('Creating Agent {0}'.format(
                 cloudify_agent['name']))
             try:
-                response = installer.runner.run_script(script_path)
-                output = response.std_out
```
isaac-s: Removed the printout block because:

1. It is incomplete. `stdout` is not enough, `stderr` is needed as well.
2. The exception will be printed by the orchestrator later anyway, so no need to repeat logging.

cloudify_agent/installer/operations.py
```diff
@@ -40,13 +41,10 @@ def create(cloudify_agent, installer, **_):
...
-                if output:
-                    for line in output.splitlines():
-                        ctx.logger.info(line)
-            except CommandExecutionError as e:
-                ctx.logger.error(str(e))
+                installer.runner.run_script(script_path)
+            except (CommandExecutionError, CommandExecutionException):
```
isaac-s: `CommandExecutionError` means that we didn't even get to execute the (potentially remote) command.
`CommandExecutionException` means that the (potentially remote) command ran, but encountered a problem.

cloudify_agent/installer/runners/winrm_runner.py
```diff
@@ -161,7 +161,8 @@ def _chk(res):
                     output=res.std_out)
                 if raise_on_failure:
                     raise error
-                self.logger.error(error)
+                self.logger.exception("WinRM command ended with an error",
```
isaac-s: Printing an `Exception` object without some sort of a prefix is a bad practice IMO

cloudify_agent/resources/script/windows-download.ps1.template
```diff
@@ -3,6 +3,9 @@
 # All output should be UTF-8 (default is UTF-16).
 $PSDefaultParameterValues['*:Encoding'] = 'utf8'
 
+# Stop whenever a non-terminating cmdlet generates an error.
+$ErrorActionPreference = "Stop"
```
isaac-s: Note that this is *not* equivalent to `bash -e`. It doesn't apply to running external programs (and `cfy-agent` is an external program).

PR #519, opened 2019-04-01
Move the rest of mgmtworker-specific code

tehasdf:


cloudify_agent/worker.py
```diff
@@ -179,23 +110,16 @@ def __init__(self, name, *args, **kwargs):
...
         kwargs = task['kwargs']
 
         logger.info(
             'Received `{0}` service task with kwargs: {1}'.format(
-                task_name, kwargs
-            )
-        )
-        result = service_tasks[task_name](**kwargs)
+                task_name, kwargs))
+        task_handler = getattr(self, service_tasks[task_name])
```
inbala: Does it work?
Shouldn’t it be `task_handler = getattr(self, self.service_tasks[task_name]`?

tehasdf: yep note that i changed 
```python
        service_tasks = {
            'ping': self.ping_task,
            'cluster-update': self.cluster_update_task,
            'cancel-workflow': self.cancel_workflow_task,
            'cancel-operation': self.cancel_operation_task
        }
```
inside of a method, to:
```python
    service_tasks = {
        'ping': 'ping_task',
        'cluster-update': 'cluster_update_task',
        'cancel-operation': 'cancel_operation_task'
    }
```
at class level, so that this can be overriden by subclasses. And so now the values of this dict are strings (with the method name) instead of just the methods (because the method objects themselves arent available then yet)

PR #520, opened 2019-04-02
Agent installer: get IPs from broker.networks, not pctx

tehasdf:


cloudify_agent/installer/config/agent_config.py
```diff
@@ -220,32 +220,15 @@ def _set_name(self):
...
-                'manager': rest_host_ip,
-                'brokers': [rest_host_ip],
-            }
-        return network
-
     def get_manager_ip(self):
-        return self._get_network()['manager']
+        managers = [manager.networks[self['network']]
+                    for manager in ctx.get_managers(network=self['network'])]
+        # TODO make the agent work with multiple managers here
```
tehasdf: won't be in this PR, let's do it in a separate task

PR #521, opened 2019-04-10
Build the cloudify-agent RPM in CI

tehasdf:


tyacbovi:
Maybe on the way update the circle ci api version

PR #530, opened 2019-04-29
Support multiple REST backends

tehasdf:


cloudify_agent/worker.py
```diff
@@ -126,6 +126,7 @@ def ping_task(self):
         return {'time': time.time()}
 
     def cluster_update_task(self, nodes):
+        # TODO! replace with the add-networks stuff
```
tehasdf: this is cy-1237

cloudify_agent/tests/test_operations.py
```diff
@@ -70,17 +71,42 @@ def _manager_env(self):
...
+        ]
+        patches = [
+            patch.dict(os.environ, new_env),
+            patch('cloudify_agent.api.utils.get_manager_file_server_url',
+                  file_server_url),
+            patch('cloudify_agent.operations._get_amqp_client',
+                  get_amqp_client),
+            patch('cloudify.endpoint.LocalEndpoint.get_managers',
+                  return_value=managers),
+        ]
```
tehasdf: added another patch so factored the patches here to avoid the 1000 nested `with`s

cloudify_agent/resources/pm/nssm/nssm.conf.template
```diff
@@ -16,10 +16,13 @@ if %errorlevel% neq 0 exit /b %errorlevel%
...
 "{{ nssm_path }}" set {{ name }} AppEnvironmentExtra ^
-REST_HOST={{ rest_host }} ^
+REST_HOST={{ rest_host|join(',') }} ^
 REST_PORT={{ rest_port }} ^
 LOCAL_REST_CERT_FILE="{{ local_rest_cert_file }}" ^
-MANAGER_FILE_SERVER_URL=https://{{ rest_host }}:{{ rest_port }}/resources ^
+MANAGER_FILE_SERVER_URL={% for host in rest_host -%}
+    https://{{ host }}:{{ rest_port }}/resources
+    {%- if not loop.last %},{% endif %}
+{%- endfor %} ^
```
tehasdf: can't do `|map|join` because jinja doesn't really support map+format. This loop is written in multiple lines, and whitespace is controlled by the `{%-`

cloudify_agent/installer/config/agent_config.py
```diff
@@ -391,7 +387,7 @@ def _set_package_url(self, runner):
 
         if agent_package_name:
             file_server_url = agent_utils.get_manager_file_server_url(
-                self['rest_host'], self['rest_port']
+                self['rest_host'][0], self['rest_port']
```
tehasdf: this (and the one below) is to be replaced with a random.choice probably. The point is, in one attempt of an installation, we will only download the package from one manager, relying on retries to try the other managers

cloudify_agent/installer/script.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, cloudify_agent):
         self.custom_env = None
         self.file_server_root = cloudify_utils.get_manager_file_server_root()
         self.file_server_url = utils.get_manager_file_server_url(
-            cloudify_agent.get_manager_ip(),
+            cloudify_agent['rest_host'][0],
```
tehasdf: ditto above

cloudify_agent/api/pm/base.py
```diff
@@ -220,15 +220,8 @@ def __init__(self, logger=None, **params):
         # Mandatory parameters
         self.validate_mandatory()
         self.rest_host = params['rest_host']
-        if isinstance(params['broker_ip'], basestring):
```
tehasdf: this now always gets a list, because the commandline param is split inside of the click parsing callback

.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ defaults:
     docker:
     # We use specifically this python image version, because there is a conflicting change in the latest python image
     # that wrongly influence them.
-      - image: circleci/python:2.7.14
+      - image: tehasdf/cloudify-agent-ci
```
tehasdf: revert

dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip
+https://github.com/cloudify-cosmo/cloudify-common/archive/multiple-rest-backends-1.zip
```
tehasdf: revert

.circleci/config.yml
geokala: Good idea!

cloudify_agent/installer/config/agent_config.py
geokala: When is this going to be replaced?
(and isn't there something in common for this?)

PR #548, opened 2019-06-02
CY-1244 Plugins install: replace local lock mechanism with DB lock,

adigrabow:
and wait for Syncthing to sync files on all Managers if cluser

### DIFF
--- a/cloudify_agent/api/plugins/installer.py
+++ b/cloudify_agent/api/plugins/installer.py
@@ -54,6 +54,7 @@ SYNCTHING_QUERY_INTERVAL = 1
 PLUGIN_QUERY_INTERVAL = 1
 INSTALLATION_TIMEOUT = 75
 
+
 class PluginInstaller(object):
 
     def __init__(self, logger=None):
@@ -153,7 +154,6 @@ class PluginInstaller(object):
                 raise exceptions.PluginInstallationError(
                     'Timeout waiting for plugin to be installed. '
                     'Plugin info: [{0}] '.format(managed_plugin))
-
         if os.path.exists(dst_dir):
             self.logger.debug('Plugin path exists {0}'.format(dst_dir))
             plugin_id_path = os.path.join(dst_dir, 'plugin.id')
@@ -175,7 +175,6 @@ class PluginInstaller(object):
                 raise exceptions.PluginInstallationError(
                     'Managed plugin installation found but it is '
                     'in a corrupted state. [{0}]'.format(managed_plugin))
-
         fields = ['package_name',
                   'package_version',
                   'supported_platform',
@@ -205,7 +204,7 @@ class PluginInstaller(object):
                     f.write(managed_plugin.id)
                 # Wait for Syncthing to sync plugin files on all managers
                 if syncthing_utils:
-                    self._wait_for_syncthing()
+                    syncthing_utils.wait_for_plugins_sync()
                 self._update_plugin_status(managed_plugin.id)
             except Exception as e:
                 tpe, value, tb = sys.exc_info()
@@ -213,19 +212,6 @@ class PluginInstaller(object):
                                           'plugin: {0} [{1}][{2}]'
                                           .format(managed_plugin.id,
                                                   plugin, e)), None, tb
-
-    @staticmethod
-    def _wait_for_syncthing():
-        deadline = time.time() + INSTALLATION_TIMEOUT
-        # This file exists only when the Syncthing service is running
-        if os.path.exists(syncthing_utils.config_path):
-            while not syncthing_utils.mgmtworker_is_plugins_syncing_complete():
-                if time.time() < deadline:
-                    time.sleep(SYNCTHING_QUERY_INTERVAL)
-                else:
-                    raise exceptions.PluginInstallationError(
-                        'Timeout waiting for Syncthing to sync plugin`s'
-                        ' directory.')
 
     @staticmethod
     def _update_plugin_status(plugin_id):
@@ -241,8 +227,7 @@ class PluginInstaller(object):
     def _is_plugin_installing(plugin_id):
         client = get_rest_client()
         plugin = client.plugins.get(plugin_id)
-        return True if plugin.archive_name.startswith(INSTALLING_PREFIX)\
-            else False
+        return plugin.archive_name.startswith(INSTALLING_PREFIX)
 
     def _wagon_install(self, plugin, args):
         client = get_rest_client()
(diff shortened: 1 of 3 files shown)


cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +140,98 @@ def _install_managed_plugin(self,
...
+        :return:
+        """
+        client = get_rest_client()
+        client.plugins.finish_installation(plugin_id)
+
+    @staticmethod
+    def _is_plugin_installing(plugin_id):
+        client = get_rest_client()
+        plugin = client.plugins.get(plugin_id)
+        if 'installing-' in plugin.archive_name:
```
tyacbovi: I think it's better to put this logic in a shared location, so does the agent will not have to impl this again

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +140,98 @@ def _install_managed_plugin(self,
...
+            return True
```
tyacbovi: This can be combined to one liner

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +140,98 @@ def _install_managed_plugin(self,
...
+    def _wait_for_syncthing():
+        # This file exists only when the Syncthing service is running
+        if os.path.exists(syncthing_utils.config_path):
+            while not syncthing_utils.mgmtworker_is_syncing_complete():
+                time.sleep(1)
+
+    @staticmethod
+    def _update_plugin_status(plugin_id):
+        """
+        Now that both the installation and Syncthing has ended the `installing`
```
tyacbovi: I will reword it to 'Completing the plugin installation process after installing it on relevant locations and syncing completion'

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +140,98 @@ def _install_managed_plugin(self,
...
-                            'does not match the ID of the plugin currently '
-                            'on the manager. [existing: {0}, new: {1}]'
-                            .format(existing_plugin_id,
-                                    managed_plugin.id))
-                else:
+
+        # create_deployment_env should wait for the install_plugin wf to finish
+        while (self._is_plugin_installing(managed_plugin.id) and
+               deployment_id != SYSTEM_DEPLOYMENT):
+            time.sleep(1)
```
tyacbovi: Better const the one second sleep, and maybe add the explanation to the connection to the syncthing loop

barucoh: magic number, perhaps export to a constant?
Also, it's worth checking if checking the DB every 1 second is a good idea.

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +140,98 @@ def _install_managed_plugin(self,
...
+                raise NonRecoverableError('Failed installing managed '
+                                          'plugin: {0} [{1}][{2}]'
+                                          .format(managed_plugin.id,
+                                                  plugin, e)), None, tb
+
+    @staticmethod
+    def _wait_for_syncthing():
+        # This file exists only when the Syncthing service is running
+        if os.path.exists(syncthing_utils.config_path):
+            while not syncthing_utils.mgmtworker_is_syncing_complete():
```
barucoh: Don't forget to update the function name here

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +140,98 @@ def _install_managed_plugin(self,
...
+                time.sleep(1)
```
barucoh: magic number, same as before

cloudify_agent/api/plugins/installer.py
barucoh: can be replaced with
`return plugin.archive_name.starts_with(INSTALL_PREFIX)`

cloudify_agent/tests/api/plugins/test_installer.py
```diff
@@ -503,7 +511,46 @@ def list(self, **kwargs):
...
+                         format(plugin_id))
+            super(ConcurrentMockPlugins, self).finish_installation(plugin_id)
+
 
 class MockClient(object):
     def __init__(self, plugins, download_path=None):
         self.plugins = MockPlugins(plugins, download_path=download_path)
+
+
+class MockConcurrentClinet(object):
```
tehasdf: s/Clinet/Client/

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +142,107 @@ def _install_managed_plugin(self,
...
+        :param plugin_id:
+        """
+        client = get_rest_client()
+        client.plugins.finish_installation(plugin_id)
+
+    @staticmethod
+    def _is_plugin_installing(plugin_id):
+        client = get_rest_client()
+        plugin = client.plugins.get(plugin_id)
+        return True if plugin.archive_name.startswith(INSTALLING_PREFIX)\
```
tehasdf: this is just `return plugin.archive_name.startswith(INSTALLING_PREFIX)` :)

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +142,107 @@ def _install_managed_plugin(self,
...
+                self._update_plugin_status(managed_plugin.id)
+            except Exception as e:
+                tpe, value, tb = sys.exc_info()
+                raise NonRecoverableError('Failed installing managed '
+                                          'plugin: {0} [{1}][{2}]'
+                                          .format(managed_plugin.id,
+                                                  plugin, e)), None, tb
+
+    @staticmethod
+    def _wait_for_syncthing():
```
tehasdf: this whole function has to do with waiting for replication so i'd put it in premium

and then this module wouldn't need to know anything about the replication impl (now it needs to know the config path and that it needs to poll)

cloudify_agent/api/plugins/installer.py
```diff
@@ -135,67 +142,107 @@ def _install_managed_plugin(self,
...
-            if lock:
-                lock.release()
+                self._wagon_install(plugin=managed_plugin, args=args)
+                shutil.move(tmp_plugin_dir, dst_dir)
+                with open(os.path.join(dst_dir, 'plugin.id'), 'w') as f:
+                    f.write(managed_plugin.id)
+                # Wait for Syncthing to sync plugin files on all managers
+                if syncthing_utils:
+                    self._wait_for_syncthing()
+                self._update_plugin_status(managed_plugin.id)
```
tehasdf: if plugin installation fails we don't set anything, unfortunate

PR #550, opened 2019-06-07
Precreate the service queue before the agent starts 

tehasdf:


tehasdf:
this also means that there will no longer _always_ be a `No response received for task b26ef5f6b80c4a8e89feecceeee595d9` message when an agent starts

cloudify_agent/api/pm/base.py
```diff
@@ -331,7 +331,17 @@ def _get_client(self):
 
     def _is_daemon_running(self):
         self._logger.debug('Checking if agent daemon is running...')
-        return utils.is_agent_alive(self.queue, self._get_client(), timeout=3)
+        client = self._get_client()
+        with client:
+            # precreate the queue that the agent will use, so that the
+            # message is already waiting for the agent when it starts up
+            client.channel_method(
```
tehasdf: yes, queue declares are idempotent

PR #553, opened 2019-06-12
linux install script: don't use `su` if not required

tehasdf:
If we're already running as the user that is supposed to install
the agent, no need to use `su`

### DIFF
--- a/cloudify_agent/resources/script/linux.sh.template
+++ b/cloudify_agent/resources/script/linux.sh.template
@@ -1,5 +1,18 @@
 #! /bin/bash -e
 
+
+MYSELF=$(whoami)
+run_as()
+{
+    user=$1
+    if [ $MYSELF == $user ]
+    then
+        ${@:2}
+    else
+        su $user --shell /bin/bash -c "set -e; ${@:2}"
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


tehasdf:
this will allow using non-ssh methods (plugin) for the sanity check, without adding cfyuser to sudoers

cloudify_agent/resources/script/linux.sh.template
```diff
@@ -1,5 +1,18 @@
...
+
+MYSELF=$(whoami)
+run_as()
+{
+    user=$1
+    if [ $MYSELF == $user ]
+    then
+        ${@:2}
+    else
+        su $user --shell /bin/bash -c "set -e; ${@:2}"
```
geokala: Why su rather than sudo?

tehasdf: because it was su in the code before the changes :P

and that used su rather than sudo because su is installed pretty much always and sudo isn't

geokala: OK- seems a bit odd because sudo is installed as part of the base image for all the versions we support, and I'm fairly sure we've got other stuff that doesn't work without it I think?

Either way, fair enough.

PR #556, opened 2019-06-24
Wait for Syncthing and update plugin name only when

adigrabow:
installing central_deployment_agent plugins


cloudify_agent/api/plugins/installer.py
```diff
@@ -202,17 +202,26 @@ def _install_managed_plugin(self,
...
             except Exception as e:
                 tpe, value, tb = sys.exc_info()
                 raise NonRecoverableError('Failed installing managed '
                                           'plugin: {0} [{1}][{2}]'
                                           .format(managed_plugin.id,
                                                   plugin, e)), None, tb
 
+    @staticmethod
+    def _is_host_agent(plugin):
+        if 'executor' in plugin:
```
tyacbovi: can one line it

PR #568, opened 2019-08-13
Set tenant in agent_config

tehasdf:
See commit messages

Instead of using utils.get_tenant in places, use ctx.tenant


dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-https://github.com/cloudify-cosmo/cloudify-common/archive/master.zip
+https://github.com/cloudify-cosmo/cloudify-common/archive/agent-config-tenant-1.zip
```
tehasdf: remember

tehasdf:
sanity check https://circleci.com/workflow-run/78ab2c87-cfcd-46e6-9d14-d19791cf447c

PR #569, opened 2019-08-13
Path args to cfy-agent: expanduser

tehasdf:
Make sure that paths like "~ubuntu/cloudify/ssl/cert.pem" are resolved.

Otherwise, click.Path tries the literal "~ubuntu" and finds that it
doesn't exist

### DIFF
--- a/cloudify_agent/shell/commands/daemons.py
+++ b/cloudify_agent/shell/commands/daemons.py
@@ -15,6 +15,7 @@
 
 import click
 import json
+import os
 
 from cloudify_agent.api import defaults
 from cloudify_agent.api import utils as api_utils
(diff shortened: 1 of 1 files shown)


cloudify_agent/shell/commands/daemons.py
```diff
@@ -125,13 +133,13 @@
...
                    .format(env.CLOUDIFY_BROKER_SSL_CERT),
               default=None,
-              type=click.Path(exists=True, readable=True, file_okay=True),
+              type=_ExpandUserPath(exists=True, readable=True, file_okay=True),
               envvar=env.CLOUDIFY_BROKER_SSL_CERT)
 @click.option('--broker-ssl-cert-path',
               help='The path to a local copy of the Broker public cert, '
                    'used for cert verification, if required [env {0}]'
               .format(env.CLOUDIFY_BROKER_SSL_CERT_PATH),
-              type=click.Path(exists=False, readable=False, file_okay=True),
```
tehasdf: `exists=False` is the default

PR #577, opened 2019-10-24
Cy 1814 plugin bundle cluster failure

glukhman:


cloudify_agent/api/plugins/installer.py
```diff
@@ -198,6 +198,10 @@ def _install_managed_plugin(self,
             try:
                 self.logger.info('Installing managed plugin: {0} [{1}]'
                                  .format(managed_plugin.id, description))
+                # Wait for Syncthing to sync resources for the wagons
+                if syncthing_utils:
+                    resources = ('resources', '/opt/manager/resources', None)
```
barucoh: This is duplicated code.
It should be pulled from `syncthing_utils` as a constant of the `resources` folder instead.

PR #578, opened 2019-10-27
CY-1852: use service_name that already exists in the daemon's configuration

isaac-s:


cloudify_agent/api/pm/base.py
```diff
@@ -708,7 +708,9 @@ def _get_script_path(self):
     def __init__(self, logger=None, **params):
         super(GenericLinuxDaemonMixin, self).__init__(logger=logger, **params)
 
-        self.service_name = 'cloudify-worker-{0}'.format(self.name)
+        self.service_name = params.get('service_name')
+        if not self.service_name:
+            self.service_name = 'cloudify-worker-{0}'.format(self.name)
```
isaac-s: @tehasdf this should really never happen, but I chose to default it to `coudify-worker-*`. Not sure if it's best. Maybe raising an exception instead?

tehasdf: could well do that too, as it is, it'll most likely throw exceptions anyway if the defaulted service name doesn't actually exist in the system. But this is fine too. I don't really see any strong reasons to go either of those.

tehasdf:
I've created a branch with the same name on cloudify-manager-install, so that CI there runs (and runs the sanity-check) using the agent built from this branch: 
https://circleci.com/workflow-run/5887afec-95e6-46fd-b319-ad74b307b6a2
(and it's green)

cloudify_agent/api/pm/base.py
isaac-s: @tehasdf I'm starting to believe that I'd prefer an exception thrown, as it implies an unexpected state. I'll push another commit shortly.

isaac-s: @tehasdf turns out that it must be conditional, I added a comment why. I'll merge after the `cloudify-manager-install` CI build proves to work.

PR #581, opened 2019-11-18
CY-1985 Make sure envvars are bytes

tehasdf:
When setting an envvar with data that is coming from json (which
is unicode) we need to make sure that the string which goes
into the environ is bytes, or otherwise we'll be having trouble
starting subprocesses with `Popen(.., env=os.environ.copy())`.

Linux subprocess/os.environ copes with values that are unicode but
only contain ascii, however windows doesn't.


bazulay-cfy:
:+1:

PR #586, opened 2019-12-03
CY-2007 replace nssm.exe with the last non-infected version (v2.21-68…

glukhman:
… x64)


glukhman:
Related ticket https://cloudifysource.atlassian.net/browse/CY-2007 is closed after communicating to the customer that the virus alert was a false positive

PR #592, opened 2019-12-26
CY-2132 add retries on download

glukhman:


cloudify_agent/api/plugins/installer.py
```diff
@@ -198,6 +198,9 @@ def _install_managed_plugin(self,
             try:
                 self.logger.info('Installing managed plugin: {0} [{1}]'
                                  .format(managed_plugin.id, description))
+                wagon_path = os.path.join('/opt/manager/resources/plugins',
```
inbala: Isn't there a constant for `/opt/manager/resources/plugins`?

PR #593, opened 2020-01-03
CY-1960 declaring queue: add x-ha-policy

tehasdf:


geokala:
Not needed, the tenant creation now copies the policies from the / vhost, and those policies apply the required configuration.

PR #603, opened 2020-03-23
CY-2360 Py3 porting: stage1

tehasdf:
Simple stuff.


.circleci/config.yml
```diff
@@ -159,11 +159,37 @@ jobs:
...
+      - checkout
+      - run:
+          name: install futurize
+          command: pip install future --user
+      - run:
+          name: find python3-incompatible code
+          command: |
+            FUTURIZE="futurize ."
+            while read line; do
+              FUTURIZE="${FUTURIZE} ${line}"
```
mateumann: (nit) maybe `-f` might be moved here from `py3fixers`

tehasdf: @mateumann i kinda prefer this way actually, allows us to put `--stage1` in fixers later, which i'm going to do i think

PR #606, opened 2020-03-25
Refactor CI, drop tox

tehasdf:


.circleci/config.yml
```diff
@@ -9,7 +9,32 @@ checkout:
...
 commands:
+  install_cron:
+    steps:
+      - run:
+          # we install cron to allow use of crontab in tests, but not run it,
+          # because we do mock cron in tests
+          name: Install and run cron
+          command: |
+                   sudo apt-get update
+                   sudo apt-get install cron
```
geokala: Probably best to do `sudo apt-get install -y cron` in case they add a dependency on cron later, which will then make this hang.

.circleci/config.yml
```diff
@@ -145,19 +124,49 @@ jobs:
           destination: result
 
   flake8:
-    <<: *test_defaults
-    environment:
-      TOX_TEST: flake8
+    docker:
+      - image: circleci/python:2.7
```
geokala: Do we not want to standardise on the same image? (2.7.14 above)

.circleci/config.yml
```diff
@@ -145,19 +124,49 @@ jobs:
...
+                   export PYENV_ROOT="$HOME/.pyenv"
+                   export PATH="$PYENV_ROOT/bin:$PATH"
+                   sudo apt-get install -y build-essential zlib1g-dev xz-utils
+                   pyenv install 2.6.9
+                   pyenv local 2.6.9
+      - run:
+          # python 2.6 will require pyyaml to be built, rather than installed
+          # from a wheel, and so we need to provide headers
+          name: Install python headers
+          command: sudo apt-get install libpython-dev
```
geokala: -y recommended again

geokala: Also, do an update first to avoid potential issues later.

.circleci/config.yml
tehasdf: no, the above things are an exception, we don't downgrade unless there's a good reason

geokala: OK, makes sense- is it worth adding a comment stating why we deliberately downgrade the above one? (I know we discussed why earlier)

PR #614, opened 2020-04-10
requirements.txt: install common without the optionals

tehasdf:
However, in the unittests, we still need them, because the unittests do
run workflows.

So this means the dev-reqs.txt, which is what is used to build the agent,
installs cloudify-common, but test-reqs.txt, which is used to run the test,
uses cloudify-common[dispatcher].

The tests then build the agent WITHOUT dispatcher, from dev-reqs.txt,
and use that in their "end-to-end" tests that install an agent


tehasdf:
commits to be renamed and the other to be reverted once cloudify-cosmo/cloudify-cosmo#456 is merged

PR #642, opened 2020-06-26
whruia

geokala:


geokala:
Oh, sorry, should've closed this one, it was an experimental alternative to the direct s3 upload.

The py2 one I think needed something fixing to actually work properly- @tehasdf might know off the top of his head, but I don't think it was trivial.

PR #654, opened 2020-07-21
CY-3094 Service task to force-install plugins

tehasdf:
And update plugin installation state.


geokala:
Appveyor fail is due to you needing to rebase.

PR #665, opened 2020-08-05
CY-3216 Delete agent AMQP resources when deleting the agent

tehasdf:
Remove that todo! Delete the agent queues and exchange.

If we fail to do that, because the agent dict is malformed - oh well,
we can't do it then.


tehasdf:
don't worry about CI, i forgot to dev-reqs, i'll rerun after cloudify-cosmo/cloudify-common#522 is merged

PR #674, opened 2020-08-13
CY-3272 Add username to mgmtworker logs

YoniItzhak:


YoniItzhak:
We add the context_manager `state.current_ctx.push(handler.ctx)` in odrer to update the current context with the `execution_creator_username`. This in turn will get to [this line ](https://github.com/cloudify-cosmo/cloudify-common/blob/427f51f105ec767f020be6c865cd0bb16b1c424e/cloudify/logs.py#L474) and get the `execution_creator_username` from the context. Before this change the context in which the function `_print_task` ran wasn't updated. 
Please let me know if you think there's a simpler solution? @geokala @tehasdf

cloudify_agent/worker.py
```diff
@@ -47,30 +47,31 @@ def __init__(self, *args, **kwargs):
...
+                suffix = ''
+            else:
+                prefix = '{0} operation'.format(action)
+                suffix = '\n\tNode ID: {0}'.format(ctx.get('node_id'))
+
+            if status:
+                suffix += '\n\tStatus: {0}'.format(status)
+
+            tenant_name = ctx.get('tenant', {}).get('name')
+            logger.info(
```
geokala, line 63: I'm wondering if it's worth changing this to use logger.info('%{thing}s', {'thing': 'something'}) approach so that the message isn't rendered if logging level is set to WARNING or above?

PR #681, opened 2020-08-18
Plugin uninstall: remove parent dir if empty

tehasdf:
Plugin dirs are `/opt/mgmtworker/env/plugins/<tenant>/<plugin>/<version>`
When the `<version>` dir is deleted, if this was the last one for that
plugin, also delete the `<plugin>` dir as well


tehasdf:
moved to cloudify-cosmo/cloudify-common#538

PR #682, opened 2020-08-18
CY-3138 Reintroduce pinning for cloudify-agent-packager

glukhman:


tehasdf:
this is because having unpinned versions in a dockerfile means docker will not rebuild the image because oh hey nothing changed in the dockerfile. Even though there's a new version published

PR #686, opened 2020-09-03
Restrict agent build location

geokala:
They must run on the same node as the fetch_rhel_images job or problems will happen.


geokala:
Closing, as @tehasdf found a better way.

PR #691, opened 2020-09-22
CY-3534 Move updating plugin state to common

tehasdf:
It now goes into the plugin installer proper.


geokala (approved):

tehasdf:
the other side of this is cloudify-cosmo/cloudify-common#555

geokala (approved):

PR #692, opened 2020-09-24
Use the AGENT_NAME envvar instead of CLOUDIFY_DAEMON_NAME

tehasdf:
We used to use both randomly, let's standardize to this one,
and forget about the other one.


tehasdf:
requires cloudify-cosmo/cloudify-common#559

geokala (approved):

geokala (approved):

PR #707, opened 2020-11-10
Fix test archive issue

yuvlar:
If the deployment of the ec2 instance failed, although there's a catchError on that stage, the post step of archiving the tests would fail the build.
Moved the post step to after the stages, and extended the sleep time for the deployment to 160 sec


yuvlar:
Opened a new relevant PR from a new branch

PR #709, opened 2020-11-11
Fix connection issue

yuvlar:
modify parts of the Jenkinsfile to have better stability while building.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -115,6 +115,7 @@ pipeline {
                       echo -e \$(cat capabilities.json | jq '.key_content.value' | tr -d '"') > ~/.ssh/ec2_ssh_key && chmod 600 ~/.ssh/ec2_ssh_key
                       sleep 120
                       ssh-keyscan -H \$(cat capabilities.json | jq '.endpoint.value' | tr -d '"') >> ~/.ssh/known_hosts
+                      echo 'ClientAliveInterval 50' | sudo tee --append /etc/ssh/sshd_config
                     """, label:'Configure and install blueprint on manager'
                   }
                 }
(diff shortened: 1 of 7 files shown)


yuvlar:
Besides modifying the Jenkinsfile to have better stability, also modified 'wait for deamon dead timeout' to 20 sec instead of 10 sec, that's because tests used to fail once in a while due to high load and timeout passed before the command, causing test to fail.

jenkins/Jenkinsfile
```diff
@@ -109,12 +109,13 @@ pipeline {
                       pip install cloudify==5.1.0
                       cfy profile use ${env.OS_IP} -u ${env.OS_USERNAME} -p ${env.OS_PASSWORD} -t ${env.OS_TENANT}
                       pushd 'bp'
-                        cfy install -b ec2-py27-blueprint ec2-py27-blueprint.yaml
+                        cfy install -b ec2-py27-blueprint-${env.BUILD_NUMBER} ec2-py27-blueprint.yaml
```
warolv, line 112: why we using a dynamic name for blueprint? it may cause a situation when you overload cloudify manager on RS with multiple blueprints and not deleting them when you finished

warolv: the question if want this?

warolv (dismissed):

jenkins/Jenkinsfile
yuvlar: It was done to prevent a clash if more than 1 build is running from the multibranch, causing the build to fail since the blueprint is already installed.
The deletion of will be done even if something fails in the stage or other stages.
That's the best way that I found, do you have any other suggestions to it?

warolv: got you, parallel builds problem

warolv (approved):

PR #710, opened 2020-11-23
Test new py3compat method

yuvlar:
Modify cloudify-agent repo to match with updated generic py3compat method


yuvlar:
--stage1 was transferred to the complete list of the set of fixes by futurize --stage1

warolv (approved):

PR #724, opened 2020-12-27
RD-862 Allow to build Centos 8

mabuaisha:
This PR added to handle a special case about cento8 generation name when it comes to download that agent on target host machine since the code-name sometimes empty and equal to the code name of Centos 7 'core'. so the generation of the name should have suffix name called `8` instead of `core` so that we can fetch the correct generated agent located inside the file server

### DIFF
--- a/cloudify_agent/resources/script/linux.sh.template
+++ b/cloudify_agent/resources/script/linux.sh.template
@@ -73,8 +73,9 @@ package_url()
     if [[ ! -z "{{ conf.package_url }}" ]]; then
         echo "{{ conf.package_url }}"
     else
-        local distro="$(python -c 'import sys, platform; sys.stdout.write(platform.dist()[0].lower())')"
-        local distro_codename="$(python -c 'import sys, platform; sys.stdout.write(platform.dist()[2].lower())')"
+        local python_bin=$(which python || which python3)
+        local distro="$($python_bin -c 'import sys, platform; sys.stdout.write(platform.dist()[0].lower())')"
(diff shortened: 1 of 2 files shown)


cloudify_agent/resources/script/linux.sh.template
```diff
@@ -73,8 +73,9 @@ package_url()
     if [[ ! -z "{{ conf.package_url }}" ]]; then
         echo "{{ conf.package_url }}"
     else
-        local distro="$(python -c 'import sys, platform; sys.stdout.write(platform.dist()[0].lower())')"
-        local distro_codename="$(python -c 'import sys, platform; sys.stdout.write(platform.dist()[2].lower())')"
+        local python_bin=$(command which python || command which python3 || echo "python")
```
geokala, line 76: Why do you use command and which here?
Wouldn't it make more sense either to use 'command -v python' or just 'which python' (and the same for python3)?

geokala: Also, defaulting back to 'python' after a lookup for the command 'python' has already fails seems like the wrong thing to do?

mabuaisha: Actually no reason for that we can use just `which python` and no need to fallback to python again

mabuaisha: @geokala I just noticed its used a lot in many places so if we go with `which` for example and `which` was not on the system would not be an issue ? which is the safer method ? I think `command -v ptyhon` ?

geokala (approved):

PR #727, opened 2020-12-28
RD-879 Use new location on windows agent upgrade

geokala:


cloudify_agent/resources/script/windows.ps1.template
```diff
@@ -54,14 +54,14 @@ function InstallAgent()
     {% if add_ssl_cert %}
     AddSSLCert
     {% endif %}
-    if (-Not (Test-Path "{{ conf.basedir }}\Scripts")) {
+    if (-Not (Test-Path "C:\Program Files\Cloudify {{ version }} Agents\Scripts")) {
```
YoniItzhak, line 57: Just a general comment. Don't we want to assign "C:\Program Files\Cloudify {{ version }} Agents" to a variable and reuse it in all places?

geokala: That seemed to be what conf.basedir was being partially used for; but it's also wrong so not useful for that. I'm checking to see if it crops up anywhere else.

geokala: As for refactoring it to have a common variable that's right: Yes, we should probably do that but I'm thinking of holding off on that for the 5.1.1 release unless I find anything to indicate that basedir being wrong will cause flames and sadness.

geokala: OK, I scrapped this bit of code, updated to a different way.

cloudify_agent/installer/config/agent_config.py
```diff
@@ -516,3 +513,10 @@ def update_agent_runtime_properties(cloudify_agent):
         cloudify_agent.pop(item, None)
     ctx.instance.runtime_properties['cloudify_agent'] = cloudify_agent
     ctx.instance.update()
+
+
+def get_windows_basedir():
+    # TODO: Get the program files directory from the machine itself
```
YoniItzhak, line 519: A wise developer once told me not to write `TODO` in my code, but rather open a JIRA about it. Which is to say, do we also have a JIRA about it?

geokala: This TODO already existed but yes, we should probably have a jira for it. I'll make a note to add one.

YoniItzhak (approved):

PR #731, opened 2020-12-30
Rd 866 cicd migrate

yuvlar:


jenkins/Jenkinsfile
```diff
@@ -203,6 +204,105 @@ EOT
...
+          options {
+            checkoutToSubdirectory('cloudify-agent')
+            timeout(time: 60, unit: 'MINUTES')
+            timestamps()
+          }
+          steps{
+            script{
+              container('shell'){
+                powershell """
+                  Write-Host "Uninstalling existing erlang installs"
```
geokala, line 231: This looks like you've just copied the agent build powershell script into here- why not download and run it instead, so that we don't have to make any fixes in both places (but by keeping the script separate it's still easy to run it by hand)?

geokala: OK, just discussed elsewhere and this was taken from the appveyor config which won't be used any more, so no objection.

geokala (approved):

PR #735, opened 2021-01-07
RD-1007 Fix issue with access home directory for agent using centos 8…

mabuaisha:
This PR added so that we can avoid any selinux permission issue raised when trying to install agent on redhat/centos 8 version where a permission denied issue appeared related to preventing systemd init process from accessing the home directory for the installed agent. where the issue happened when starting the agent daemon [start daemon](https://github.com/cloudify-cosmo/cloudify-agent/blob/master/cloudify_agent/resources/script/linux.sh.template#L166)  where it configure the systemd required files for agent so that it can be controlled and running with systemctl and run when machine get rebooted.

The starter command for that service will try to start the host agent that located under `/home/centos` or `/home/cloud-user` directory and when this try to get started it failed because of permission issues raised by selinux.

### DIFF
--- a/cloudify_agent/resources/script/linux-download.sh.template
+++ b/cloudify_agent/resources/script/linux-download.sh.template
@@ -8,11 +8,27 @@ run_as()
     user=$1
     if [ "$MYSELF" == "$user" ]
     then
-        ${@:2}
+        sg cfyagent "${@:2}"
     else
         su $user --shell /bin/bash -c "set -e; ${@:2}"
     fi
 }
+
+configure_agent_installation()
+{
+   # Create cfyagent group so that we can assign it user that execute and install agent
+   # to that group so that they can have access to `/opt/cloudify-agent-{version}`
+   user_groups="$(groups {{ user }})"
+   grep -q -e 'cfyagent' /etc/group || sudo groupadd cfyagent
+   grep -q -e 'cfyagent' <<< "$user_groups" || sudo usermod -a -G cfyagent {{ user }}
+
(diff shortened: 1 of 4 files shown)


cloudify_agent/resources/script/linux-download.sh.template
```diff
@@ -17,7 +17,8 @@ run_as()
 add_ssl_cert()
 {
     # Create all the directories in the path to the cert file
-    mkdir -p $(dirname {{ ssl_cert_path }})
+    sudo mkdir -p $(dirname {{ ssl_cert_path }})
+    sudo chown -R {{ user }}. $(dirname {{ agent_dir }})
```
geokala, line 21: It probably shouldn't be owned by the user, but should be readable by them- perhaps we should add a cloudify-agent group and add the current user to it?
Otherwise, aside from any best practices, we'd have an issue if two different users had agents deployed on the same server.

mabuaisha: Agree we should allow any user execute and install agent to have access to `/opt/cloudify-agent-{version}` I missed that point will check it

geokala (approved):

geokala (changes requested):
We can avoid trying to string-match the group if we instead do:
sudo groupadd -f cfyagent
sudo usermod -a -G cfyagent {{ user }}

Both of these will return success if they make the requested change or don't need to do so (as far as I have seen in a quick test).

geokala (approved):

PR #743, opened 2021-01-28
RD-1187 Delete queue and exchange resources on agent delete for detac…

mabuaisha:
This PR created in order to address the issue of remaining exchanges & queues for agent when it runs with plugin installation method using detach mode. The issue is that when runs `cfy-agent daemons delete` it will try to connect to rabbitmq again and declare exchanges/queues again  that cause to be there after the agent get deleted.

### DIFF
--- a/cloudify_agent/api/pm/detach.py
+++ b/cloudify_agent/api/pm/detach.py
@@ -64,11 +64,55 @@ class DetachedDaemon(CronRespawnDaemonMixin):
         self._runner.run(self.create_disable_cron_script())
         super(DetachedDaemon, self).before_self_stop()
 
+    def _delete_queue(self, client):
+        queue = '{0}_service'.format(self.queue)
+        self._logger.info('Deleting amqp agent queue %s', queue)
+        try:
+            client.channel_method(
+                'queue_delete',
+                queue=queue,
+                if_empty=True,
+                wait=True
+            )
+        except Exception as err:
+            self._logger.warning('Error deleting queue'
+                                 ' queue %s: %s', queue, err)
+        else:
+            self._logger.info(
+                'AMQP agent queue %s is deleted '
+                'successfully', queue)
+
+    def _delete_exchange(self, client):
+        self._logger.info('Deleting amqp agent exchange %s', self.queue)
+        try:
+            client.channel_method(
+                'exchange_delete',
+                exchange=self.queue,
+            )
+        except Exception as err:
+            self._logger.warning('Error deleting exchange '
+                                 '%s: %s', self.queue, err)
+        else:
+            self._logger.info(
+                'AMQP agent exchange %s is deleted '
+                'successfully', self.queue)
+
+    def delete_agent_resources(self):
+        self._logger.info('Deleting amqp agent resources...')
+        client = self._get_client()
+        with client:
+            self._delete_queue(client)
+            self._delete_exchange(client)
+        self._logger.info('All amqp agent resources are deleted '
+                          'successfully')
+
     def delete(self, force=defaults.DAEMON_FORCE_DELETE):
         if self._is_daemon_running():
             if not force:
                 raise exceptions.DaemonStillRunningException(self.name)
             self.stop()
+
+        self.delete_agent_resources()
 
         if os.path.exists(self.pid_file):
             self._logger.debug('Removing {0}'.format(self.pid_file))


cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
...
+                self._logger.debug(
+                    'AMQP agent exchange {0} deleted '
+                    'successfully'.format(self.queue)
+                )
+
+    def clean_agent_resources(self):
+        self._logger.debug('Cleaning remaining amqp agent resources...')
+        client = self._get_client()
+        self._clean_queue(client)
+        self._clean_exchange(client)
```
tehasdf, line 113: use a single `with client` for both, no need to do multiple connections

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
         self._runner.run(self.create_disable_cron_script())
         super(DetachedDaemon, self).before_self_stop()
 
+    def _clean_queue(self, client):
+        queue = '{0}_service'.format(self.queue)
+        self._logger.debug('Cleaning amqp agent queue {0}'.format(queue))
```
tehasdf, line 69:
```suggestion
        self._logger.debug('Cleaning amqp agent queue %s', queue)
```
as always, the usual comment about str.format and logging
(no need to format the string if it's not going to be emitted, like in case of debug logs)

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
         self._runner.run(self.create_disable_cron_script())
         super(DetachedDaemon, self).before_self_stop()
 
+    def _clean_queue(self, client):
```
tehasdf, line 67:
```suggestion
    def _delete_queue(self, client, queue_name):
```
it's not _cleaning_ the queue, it's deleting it

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
...
+                self._logger.debug(
+                    'Either queue {0} does not exist or '
+                    'something else happened {1}'.format(queue, str(err)))
+            else:
+                self._logger.debug(
+                    'AMQP agent queue {0} deleted '
+                    'successfully'.format(queue)
+                )
+
+    def _clean_exchange(self, client):
```
tehasdf, line 88:
```suggestion
    def _delete_exchange(self, client):
```
ditto

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
...
+        self._clean_queue(client)
+        self._clean_exchange(client)
+
     def delete(self, force=defaults.DAEMON_FORCE_DELETE):
         if self._is_daemon_running():
             if not force:
                 raise exceptions.DaemonStillRunningException(self.name)
             self.stop()
+        # Make sure that all queues/exchanges get deleted to avoid
+        # leaked amqp resources
```
tehasdf, line 121: no need to comment things that are obvious from the code, `i += 1  # increment i`

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
...
+            ' exchange {0}'.format(self.queue)
+        )
+        with client:
+            try:
+                client.channel_method(
+                    'exchange_delete',
+                    exchange=self.queue,
+                )
+            except Exception as err:
+                self._logger.debug(
```
tehasdf, line 100: higher than debug

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
...
+                    'Either exchange {0} does not exist or '
+                    'something else happened {1}'.format(self.queue, str(err)))
```
tehasdf, line 102: no need for str(), it already is stringified

tehasdf, line 102: "something else happened" lol!
```suggestion
                    'Error deleting exchange %s: %s', self.queue, err)
```

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
...
+    def _clean_queue(self, client):
+        queue = '{0}_service'.format(self.queue)
+        self._logger.debug('Cleaning amqp agent queue {0}'.format(queue))
+        with client:
+            try:
+                client.channel_method(
+                    'queue_delete',
+                    queue=queue,
+                    if_empty=True,
+                    wait=False
```
tehasdf, line 76: why no wait? we want to see the response for sure

cloudify_agent/api/pm/detach.py
```diff
@@ -64,11 +64,62 @@ def before_self_stop(self):
...
+                    'Either exchange {0} does not exist or '
+                    'something else happened {1}'.format(self.queue, str(err)))
+            else:
+                self._logger.debug(
+                    'AMQP agent exchange {0} deleted '
+                    'successfully'.format(self.queue)
+                )
+
+    def clean_agent_resources(self):
+        self._logger.debug('Cleaning remaining amqp agent resources...')
```
tehasdf, line 110: i don't see how they are "remaining", obviously we _just_ created them

tehasdf (commented):
in general i suppose it makes sense to clean up after ourselves in a `delete` call, if we are to indeed create the queue/exchange, we should delete them after we're done indeed

tehasdf (approved):

PR #757, opened 2021-02-10
Add chmod to other linux install template

geokala:


cloudify_agent/resources/script/linux.sh.template
```diff
@@ -26,8 +26,13 @@ configure_agent_installation()
...
    sudo mkdir -p {{ conf.agent_dir }}
-   sudo chgrp -R cfyagent $(dirname {{ conf.agent_dir }})
-   sudo chmod -R g+w $(dirname {{ conf.agent_dir }})
+   # Currently agent data is expected to live in ~user/<agent name>, so we need to allow cfyagent group access to there,
+   # but without breaking everything else in the homedir (e.g. we shouldn't mess with permissions in .ssh)
+   sudo chgrp cfyagent $(dirname {{ agent_dir }})
+   sudo chmod g+x $(dirname {{ agent_dir }})
+   # Ensure that the cfyagent user can read and write and files and directories, and access any dirs
+   sudo chgrp -R cfyagent {{ agent_dir }}
+   sudo chmod -R g+rwX {{ agent_dir }}
```
mabuaisha, line 35: @geokala you missed the `conf.` this is different from the other place for linux-download.sh

mabuaisha (changes requested):
Left small comment

mabuaisha (approved):

mabuaisha (dismissed):

cloudify_agent/api/pm/base.py
```diff
@@ -485,6 +485,7 @@ def stop(self,
         shutdown.
 
         """
+        self.before_self_stop()
```
tehasdf, line 488: we're adding `before_self_stop` to `stop` itself, because it must _always_ be called together. It used to be that real usage did remember to call it, but tests forgot. So, we're making just `stop` call it so that the caller doesn't need to remember.

tehasdf (approved):

PR #759, opened 2021-02-11
RD-1218 Separate packaging files for RHEL6 (#745)

mabuaisha:
This contains 5 commits related to the build itself which solved the following issues in case we need to re-build the 5.1.2 again:
1. Fix issue with pip version for py2 as the pipeline for 5.1.2 is still referenec the latest pip version which will be 21.x
2. Fix issue with running tests using py2
3. Add windows pipeline to the 5.1.2 since it was used the appvyor before
4. Fix issue with redhat 6 build as the 5.1.2 build still reference the old one


jenkins/Jenkinsfile
```diff
@@ -1,19 +1,18 @@
...
+    [envVar: 'MANAGER_IP', vaultKey: 'ip'],
+    [envVar: 'MANAGER_PASSWORD', vaultKey: 'password']]],
 ]
 
 @Library('pipeline-shared-library') _
 
 pipeline {
   agent {
     kubernetes {
-      label "cloudify-agent-${env.BUILD_NUMBER}"
```
geokala, line 16: I thought this was added to work around a problem- why is it removed for his?

jenkins/Jenkinsfile
```diff
@@ -28,10 +27,12 @@ pipeline {
     PROJECT = "cloudify-agent"
     VERSION = getVersion("${env.PROJECT}", "${env.BRANCH_NAME}").trim()
     PRERELEASE = getPreRelease("${env.PROJECT}", "${env.BRANCH_NAME}").trim()
-    PATH = "/root/.local/bin:$PATH"
   }
   stages{
     stage ('compatability and flake8') {
+      environment {
+        PATH = "/root/.local/bin:$PATH"
```
geokala, line 34: Why does this need to be moved as part of the RHEL6 packaging?

jenkins/Jenkinsfile
```diff
@@ -203,6 +209,105 @@ EOT
...
+                  if ("\$status" -ne "Running") {
+                    Write-Host "WARNING: RabbitMQ DID NOT START"
+                    \$host.SetShouldExit(1)
+                  }
+                """
+                echo 'Install python'
+                powershell """
+                  echo "### Install Chocolatey ###"
+                  Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
+                  choco install -y python --version=3.6.0
```
geokala, line 282: We are supposed to be using embedded python for this- why are we using choco to install it, and using a different version than we test with (embedded version was I thought 3.6.8 or something? would have to check with an installed agent).

mabuaisha: @geokala I cherry-picked this from @yuvlar commit into master not sure why he did this wat

jenkins/Jenkinsfile
```diff
@@ -203,6 +209,105 @@ EOT
...
+                  choco install -y git
+                """
+                echo 'Configure python'
+                powershell """
+                  SET \$PATH=${env.PYTHON};\$PATH
```
geokala, line 287: If we do the install using either the embedded approach we use for the agent itself or using a venv then we don't need to mess around with paths like this.

jenkins/Jenkinsfile
mabuaisha: There were some changes happened recently regadring that and I cherry-picked all the related commit for the pipelines

@yuvlar any idea why you did it like this ?

jenkins/Jenkinsfile
```diff
@@ -203,6 +209,105 @@ EOT
...
+                  python --version
+                  python -c "import struct; print(struct.calcsize('P') * 8)"
+                  echo Upgrading pip...
+                  python -m pip install --upgrade pip==9.0.1
+                  pip --version
+                """
+                powershell """
+                  git clone https://github.com/cloudify-cosmo/cloudify-agent.git C:\\projects\\cloudify-agent
```
geokala, line 295: If we just retrieve the archive of the branch directly then we don't need to play with git- is there any good reason for us to involve git in this?

jenkins/Jenkinsfile
mabuaisha: @yuvlar since you update the windows pipelines any ideas why you did this ?

geokala: In fact, I see we do use a venv later... so this seems like it should be unnecessary?

jenkins/Jenkinsfile
mabuaisha: CC @yuvlar any reason for that this is cherry picked from your changes from master

jenkins/Jenkinsfile
mabuaisha: @yuvlar I think you did it for a reason right ?

yuvlar: because this method is deprecated, 
when not specifying a label, it'll automatically create a random label to the agent, with the project name, branch and build number

jenkins/Jenkinsfile
yuvlar: the py3compat stage needs that to be added to its PATH.
in this pipeline it uses a shared method that doesn't include this PATH, so I added it before.

jenkins/Jenkinsfile
yuvlar: I simply moved the code from AppVeyor, to use it I installed choco.
If there's any other preferred method, I don't mind changing it

jenkins/Jenkinsfile
yuvlar: again, pasted it from AppVeyor, which was working fine.
It passed a review by Igor and it passes in the CI.
If there's an issue with it let's open a Jira and I'll take care of it

jenkins/Jenkinsfile
yuvlar: Same, like comments above

PR #775, opened 2021-03-24
Clean up existing agent before install

geokala:


geokala:
This is not where the latest problem lies, closing.

PR #777, opened 2021-04-02
test-reqs: pin decorator to 4.4.2

tehasdf:
otherwise networkx will pull in 5.0.0, which is incompatible with py2.
Anyway, this only matters for the tests, because the actual agent
doesn't use networkx


geokala (dismissed):

jenkins/Jenkinsfile
```diff
@@ -315,15 +313,15 @@ EOT
...
     }
-    stage('Terminate py27 instnace'){
+    stage('Terminate instance'){
       steps{
-        container('py27'){
+        container('py36'){
           dir("${env.WORKSPACE}/${env.PROJECT}/jenkins"){
             withVault([configuration: configuration, vaultSecrets: secrets]){
               echo 'Uninstall and delete blueprint from manager'
               sh """#!/bin/bash
```
geokala, line 322: This probably wants the set incantation too- more than one line of script

geokala (approved):

geokala (approved):

PR #794, opened 2021-05-31
increase resources

yuvlar:


mabuaisha (dismissed):

jenkins/Jenkinsfile
```diff
@@ -230,96 +233,99 @@ EOT
...
-                  Write-Host "Installing RabbitMQ..."
-                  # Don't wait for rabbitmq installer as it seems to never return when run silently
-                  Start-Process "${env.RABBITMQ_INSTALLER_PATH}" -ArgumentList /S,/QN -NoNewWindow
-                  for (\$attempt=1; \$attempt -le 60; \$attempt=\$attempt+1) {
-                    \$status=(Get-Service RabbitMQ).Status
-                    if ("\$status" -eq "Running") {
-                      Write-Host "RabbitMQ is running."
-                      Break
+            catchError(message: 'Failure on: cloudify-agent tests', buildResult: 'SUCCESS', stageResult: 'FAILURE') {
+              script{
```
warolv, line 237: do we need 'script {' ?

warolv (changes requested):

jenkins/Jenkinsfile
yuvlar: I didn't really dig into it, that was working like that before

warolv: ok, so let's remove it, if no need in script{

yuvlar: I'll try to run this pipeline without it and update

warolv (approved):

PR #799, opened 2021-06-09
RD-2586 Ec2 on spot

yuvlar:
RD-2586, RD-2587
Update ec2 to use spot, and move pod to spot instance type instead of spot-xlarge


geokala:
This has been open for more than six months so I'm closing it. If it's still wanted, please reopen it and ping me so we can get it reviewed.

PR #825, opened 2022-01-10
RD-3946 Don't leak bypass_maintenance between executions

tehasdf:
Obviously, if we set it in the environ, and never set it back, that's
no bueno.

Funnily, we don't even need it here, because in the current process,
it's already set on ctx; and we pass it to child processes explicitly,
when creating _their_ env.


tehasdf:
the pynacl one is unrelated, but we need to do it, to be able to still pass CI (see description)

geokala (approved):

PR #831, opened 2022-04-17
RD-2586 use spot fleet

yuvlar:
After a long time:
First CI pipeline to use a cloudify blueprint that utilizes spot-fleet feature.
Updated BP to use spot fleet with pre-created roles, profiles and infra resources.

updated jenkinsfile with latest changes.

After this is quiet for a while, implementation will occur on other pipelines as well.


bartoszkosciug (commented):
looks okay generally please just see my comment regarding naming of the stage step

jenkins/Jenkinsfile
```diff
@@ -336,15 +334,17 @@ EOT
     }
     stage('Terminate py27 instnace'){ // we made github to wait on a job with this exact name, typo included
```
bartoszkosciug, line 335: Shouldn't we change the py27 to py36?

yuvlar: yes indeed.
Change incoming in a minute

yuvlar: FYI, in such cases we also need to update the branch protection to match the stage name.

bartoszkosciug (dismissed):
looks very good!

bartoszkosciug (approved):

PR #837, opened 2022-05-23
solid types 6.4.0-dev - 6.3.0.

bartoszkosciug:


yuvlar (approved):

geokala:
Given that this pin is now at least a version behind the current stable, I'm closing this. If we do want something similar, please reopen with 6.4.0 pin instead of 6.3.0 (unless there is some pressing reason to use 6.3.0).

PR #841, opened 2022-08-07
RD-5553 Windows jnlp agent connectivity ci

yuvlar:
A known issue in Jenkins causes windows JNLP containers to take a few seconds to connect.
Adding the args to retry the connection after a few seconds seem to stabilize it.

Also using an older types.yaml in the blueprint to have it working with the 1_3 dsl as the manager+cli used is prior to 6.4.0


yuvlar:
Checks won't pass as there are issues with master tests

tehasdf (dismissed):

tehasdf (approved):

PR #843, opened 2022-08-26
Bump jinja2 version to the same as in cloudify-common

tehasdf:


tehasdf:
RD-5692 will partially revert this

geokala (approved):

PR #845, opened 2022-09-12
RD-4646 rewrite cloudify_agent.installer.operations.start

tehasdf:
This operation will now not use retries, but rather
in-operation polling. This way, setting manager task retries
to a low number, won't break agent installations. Also, polling
could then be more responsive (more often) and with a given
timeout as well.

FWIW, the timeout was 15 before.


cloudify_agent/installer/operations.py
```diff
@@ -81,33 +88,66 @@ def configure(cloudify_agent, installer, **_):
...
-            update_agent_record(cloudify_agent, AgentState.NONRESPONSIVE)
-        return ctx.operation.retry(
-            message='Waiting for Agent to start...')
+    marked_nonresponsive = False
+    # we'll wait up to an hour for the agent to start. It really should
+    # start sooner than that, but we have to give up eventually.
+    start_time = time.time()
+    deadline = start_time + 3600
+    # we can't poll too often, because we need to give the agent a chance to
+    # respond. 10 seconds should be PLENTY even on very latency-deficient
```
geokala, line 112: I'll just leave this here... https://web.archive.org/web/20220330010954/https://www.blug.linux.no/rfc1149/writeup/

But yeah, probably we can go a bit less pathological!

geokala (dismissed):

cloudify_agent/installer/operations.py
tehasdf: used to be 15 before! I think 10 is also nice.

geokala (approved):

PR #847, opened 2022-09-12
Rd 5909 use py310 tgz

glukhman:


glukhman:
Closing as we're probably move away from cloudify-agent-packager (cfy-ap) altogether

PR #849, opened 2022-09-14
Rd 5919 agents to py310

glukhman:


cloudify_agent/shell/commands/configure.py
```diff
@@ -49,8 +45,63 @@ def configure(disable_requiretty, relocated_env, no_sudo):
...
         _disable_requiretty(no_sudo)
         click.echo('Successfully disabled requiretty for cfy-agent')
-    if relocated_env:
-        _relocate_venv()
+    if fix_shebangs:
+        click.echo('Fixing shebangs in scripts')
+        _fixup_scripts()
+
+
+def _fixup_scripts():
```
tehasdf, line 53: I'll note here for the future reader (probably me in 6 months): this is inspired by #107

cloudify_agent/resources/script/linux.sh.template
```diff
@@ -137,10 +137,10 @@ export -f install_agent
 configure_virtualenv()
```
tehasdf, line 137: I had also hoped we'd be able to shorten this script quite a bit. Maybe just inline this function? No need to keep talking about a "virtualenv"

cloudify_agent/resources/script/linux.sh.template
```diff
@@ -137,10 +137,10 @@ export -f install_agent
...
 {
     export_daemon_env
-    # configure command is run explictly as the virtualenv has not been "fixed"
-    # yet
+    # configure command is run explictly as the shebangs in the "virtualenv"
+    # have not been "fixed" yet
     echo "Configuring virtualenv..."
-    ${AGENT_PYTHON_INTERPRETER} ${AGENT_CLI} configure --relocated-env
+    sudo ${AGENT_PYTHON_INTERPRETER} ${AGENT_CLI} configure --fix-shebangs
```
tehasdf, line 143: ?! this can't require sudo, more, this function is explicitly run with `run_as "{{ conf.user }}" configure_virtualenv`

cloudify_agent/shell/commands/configure.py
```diff
@@ -49,8 +45,63 @@ def configure(disable_requiretty, relocated_env, no_sudo):
...
+    """Make scripts in bin_dir relative by rewriting their shebangs
+
+    Examine each file in bin_dir - if it looks like a python script, and has a
+    shebang - replace it with a new, "relative" shebang.
```
geokala, line 57: Why is relative in quotes here? If it is relative, surely we can do it at build time. If it's not relative, can we please not use that term here or in the logger.debug below.

cloudify_agent/shell/commands/configure.py
```diff
@@ -49,8 +45,63 @@ def configure(disable_requiretty, relocated_env, no_sudo):
...
+    """Search bin_dir for files that look like python scripts with a shebang
+    """
+    for filename in os.listdir(bin_dir):
+        filename = os.path.join(bin_dir, filename)
+        if not os.path.isfile(filename):   # ignore subdirs, e.g. .svn ones
+            continue
+
+        with open(filename, 'rb') as f:
+            try:
+                lines = f.read().decode('utf-8').splitlines()
```
geokala, line 79: This isn't great. For every file we're investigating, we read it in entirely, and then later we read it entirely again.

I'd suggest we either combine the functions so that we read, then immediately rewrite if we should do so; or we modify this so that we only read the first, say, 25 characters.
However, I think rolling the two into one would make more sense- no point opening the file twice.

cloudify_agent/shell/commands/configure.py
```diff
@@ -49,8 +45,63 @@ def configure(disable_requiretty, relocated_env, no_sudo):
...
+            except UnicodeDecodeError:
+                # This is probably a binary program instead
+                # of a script, so just ignore it.
+                continue
+
+        if not lines:
+            continue
+
+        shebang = lines[0]
+        if not (shebang.startswith('#!') and 'bin/python' in shebang):
```
geokala, line 89: This is broken by things like: #! /usr/bin/env python
I'm not sure whether we do care about that, though?

jenkins/Jenkinsfile
```diff
@@ -160,30 +139,35 @@ EOT
...
-              """, label: 'Copying repo to separate workspace'
+                git clone https://github.com/cloudify-cosmo/cloudify-agent.git SOURCES && cd SOURCES && git checkout ${BRANCH_NAME}
+              ''', label: 'Copying repo to separate workspace'
               sh script: """
                 curl -o ~/rpmbuild/SOURCES/cloudify-agents.spec https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager/master/packaging/cloudify-agents.spec ;
                 mkdir -p ~/rpmbuild/SOURCES/packaging/agents/ ;
                 curl -o ~/rpmbuild/SOURCES/packaging/agents/copy_packages.py https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager/master/packaging/agents/copy_packages.py ;
                 """, label: 'Prepare rpm-building environment'
-              sh script: """
+              sh script: '''
```
geokala, line 151: This needs at the very least a "set -e" so that it doesn't continue after something fails. As noted in other comments, it may well want set -eux, and as noted elsewhere it may want to set the pipefail option (so that we have one standard invocation everywhere).

jenkins/Jenkinsfile
```diff
@@ -160,30 +139,35 @@ EOT
...
                 cd ~/rpmbuild/SOURCES
-                cp -rf ${env.WORKSPACE}/build_agent/*.tar.gz .
+
+                #--- build the agent here
+                cp /rpmbuild/SOURCES/packaging/linux_agent_builder.sh .
+                chmod +x linux_agent_builder.sh
+                ./linux_agent_builder.sh centos core ${VERSION}-${PRERELEASE}
+
                 for i in {1..30}; do yum-builddep -y cloudify-agents.spec && break || sleep 10; done
```
geokala, line 159: While we're here can we add a comment noting that we do this up to 30 times because yum mirrors aren't always entirely reliable?
I think I put a bunch of these retries in various places, so should've commented them earlier, sorry!

packaging/linux/linux_agent_builder.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+set -eux
```
geokala, line 2: We don't technically need it yet, but it might be worth adding the pipefail option to this so we have just one standard invocation that we can copy-paste everywhere and get well behaved scripts.

packaging/linux/linux_agent_builder.sh
```diff
@@ -0,0 +1,32 @@
...
+
+DISTRO="${1:-centos}"
+RELEASE="${2:-core}"
+CFY_VERSION="$3"
+
+AGT_NAME="$DISTRO-$RELEASE-agent_$CFY_VERSION"
+WORKDIR="$(pwd)"
```
geokala, line 9: Why are we setting this as a variable- and one we don't use, at that?

packaging/linux/linux_agent_builder.sh
```diff
@@ -0,0 +1,32 @@
...
+
+mkdir -p cloudify/env
+
+# This is going to create an agent .tar.gz with a python inside the agent's
+#   env/ folder, not a true virtualenv.
+
+# -- copy python 3.10 to "venv"
+curl https://cloudify-cicd.s3.amazonaws.com/python-build-packages/cfy-python3.10.tgz -o cfy-python3.10.tgz
```
geokala, line 17: We could use "--retry 5" (or some other value) on this if we think there might be occasional connection issues with s3 (which is not an unreasonable assumption).

packaging/linux/linux_agent_builder.sh
```diff
@@ -0,0 +1,32 @@
...
+pushd 'cloudify/env/bin'
+    ln -s python3.10 python
+    ln -s python3.10 python3
+    ln -s pip3.10 pip
+    ln -s pip3.10 pip3
+popd
+
+# -- install agent
+cloudify/env/bin/python3.10 -m pip install setuptools upgrade
+cloudify/env/bin/python3.10 -m pip install -r dev-requirements.txt
```
geokala, line 28: Installing dev-requirements for something that is part of the (eventual) production pipeline? That seems wrong. Have we got a badly named reqs file?

packaging/linux/linux_agent_builder.sh
geokala: It'd also be really nice if we looked at putting at least some hashes to check to avoid downloading a broken package and only finding out later, but I think that's probably for later.

packaging/linux/linux_agent_builder.sh
glukhman: this actually installs cloudify-common, so maybe just rename the file requirements.txt ?

cloudify_agent/shell/commands/configure.py
tehasdf: it's not broken: `#!/usr/bin/env python` would be correct and wouldn't need changing by this (because it's not a hardcoded path to the python executable, and this code aims to fixup hardcoded paths to python executables).

Also, those files never contain `env python`, but they always contain hardcoded paths, and they pretty much have to: they need to run the virtualenv (or "virtualenv") python, even without being activated first

jenkins/Jenkinsfile
```diff
@@ -160,30 +139,37 @@ EOT
...
+              ''', label: 'Copying repo to separate workspace'
+              sh script: '''
                 curl -o ~/rpmbuild/SOURCES/cloudify-agents.spec https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager/master/packaging/cloudify-agents.spec ;
                 mkdir -p ~/rpmbuild/SOURCES/packaging/agents/ ;
                 curl -o ~/rpmbuild/SOURCES/packaging/agents/copy_packages.py https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager/master/packaging/agents/copy_packages.py ;
-                """, label: 'Prepare rpm-building environment'
-              sh script: """
+                ''', label: 'Prepare rpm-building environment'
+              sh script: '''
+                set -euxo pipefail
```
tehasdf, line 152:
```suggestion
```
cf [my other comment](https://github.com/cloudify-cosmo/cloudify-agent/pull/849/files#r974574140)

packaging/linux/linux_agent_builder.sh
tehasdf: As noted on the channel, curl is really stupid about return codes: it'll happily return 0 on errors. It needs `-f` to actually abort on failures.
So I suggest this:
```suggestion
curl -f --retry 5 https://cloudify-cicd.s3.amazonaws.com/python-build-packages/cfy-python3.10.tgz -o cfy-python3.10.tgz
```

packaging/linux/linux_agent_builder.sh
tehasdf: Let's do the very-correct thing, with `reqs.in` and `reqs.txt` and `pip-compile`, but... not right now. I'd like to do this as part of also relaxing pins in setup.py (the same way we did in restservice etc).

So: yes, but please not now.

cloudify_agent/shell/commands/configure.py
tehasdf: it's now doing `.readline()`. Good.
BTW, I note that shebangs are effectively very limited size (128, by linux kernel hardcoding), and file reads work in terms of disk pages (well... 4096?), so 25 is unnecessarily limiting :)

cloudify_agent/shell/commands/configure.py
```diff
@@ -49,8 +45,58 @@ def configure(disable_requiretty, relocated_env, no_sudo):
...
+def _fixup_scripts():
+    """Make scripts in bin_dir relative by rewriting their shebangs
+
+    Examine each file in bin_dir - if it looks like a python script, and has a
+    shebang - replace it with a shebang pointing to the agent's python.
+    """
+    from cloudify_agent.shell.main import get_logger
+    logger = get_logger()
+
+    new_shebang = f"#!{sys.executable}"
```
tehasdf, line 62: we need to understand what `sys.executable` actually is. If it's `/opt/cloudify-agent-7.0.0/bin/python`, fine. But if it contains the agent node id in its path, then I don't think we can do this. Shebangs are limited to 128 bytes. This is tested by the "long path" test in this repo, so if that keeps failing, this would be why. Maybe we'll need something like virtualenv's relocatable. That'd be unfortunate.

tehasdf: OK. We did the silly "multi-line shebang" trick

tehasdf (approved):

PR #853, opened 2022-09-19
RD-5982 Use `distro` module to find Linux distribution

mateumann:


tehasdf (approved):

tehasdf:
I've pulled this one into #854, going into #849, because there's no way of this passing separately, or that other one passing without this

PR #859, opened 2022-10-06
Bump click to match manager

glukhman:


setup.py
```diff
@@ -20,7 +20,7 @@
     'appdirs==1.4.3',
     'jinja2>3,<4',
     'virtualenv==15.1.0',
-    'click>7,<8',
+    'click>8,<9',
```
geokala, line 23: Worth noting that this only fixes it until flask update their deps.

geokala (approved):

PR #867, opened 2023-01-23
RD-6658 Use `cfy-agent setup` for creating an agent

tehasdf:
This PR is currently just a single draft WIP commit, but it'll be rewritten soon.


cloudify_agent/api/pm/base.py
```diff
@@ -168,108 +161,88 @@ class Daemon(object):
...
+        logger=None,
+        broker_ip,
+        name,
+        queue,
+        deployment_id=None,
+        user=None,
+        broker_user='guest',
+        broker_pass='guest',
+        broker_vhost='/',
+        broker_ssl_enabled=False,
```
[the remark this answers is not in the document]
tehasdf: They always are... .except in this CI

cloudify_agent/api/pm/base.py
```diff
@@ -168,108 +161,88 @@ class Daemon(object):
...
-
-        # will be populated later on with runtime properties of the host
-        # node instance this agent is dedicated for (if needed)
-        self._runtime_properties = None
-
-        # configure logger
+    def __init__(
+        self,
+        *,
+        logger=None,
```
geokala, line 165: Might also be worth type-hinting them?

geokala: (with the validation removed that'd have some more benefit I suspect)

cloudify_agent/resources/script/linux.sh.template
```diff
@@ -1,236 +1,273 @@
-#! /bin/bash -ex
+#!/bin/bash -e
```
geokala, line 1: No point having -e here with the set -euo later

cloudify_agent/resources/script/linux.sh.template
```diff
@@ -1,236 +1,273 @@
...
+
+    log "Agent started successfully!"
 }
-export -f main
-{% if tmpdir %}
-export TMPDIR="{{ tmpdir }}"
-{% endif %}
-main
+
+main 2>&1 | run_as_root tee "$LOG_FILE"
```
geokala, line 273: I was wondering where you were doing something like this. Nice.

cloudify_agent/installer/script.py
```diff
@@ -226,13 +168,12 @@ def install_script_download_link(cloudify_agent=None, **_):
...
+    # back-compat
     script_builder = _get_script_builder(cloudify_agent=cloudify_agent)
-    return script_builder.init_script()
+    return script_builder.install_script()
 
 
-def init_script_download_link(cloudify_agent=None, **_):
-    script_builder = _get_script_builder(cloudify_agent=cloudify_agent)
-    return script_builder.init_script_download_link()
+install_script = init_script
```
geokala, line 176: Doesn't this entirely shadow install_script?

cloudify_agent/api/pm/base.py
tehasdf: eh, not that useful. It's all strings...

cloudify_agent/installer/script.py
tehasdf: we used to have both `init_script` and `install_script` here, which emitted the linux.sh/linux-download.sh - both eventually did the same thing.

I keep both functions on this level for some "back-compat" - we don't need to, but eh, why not. This is mostly only called from `ctx.plugin.init_script()` (which calls `.init_script` from here)

Anyway, shadow - not really, there's no other function named `install_script` here

cloudify_agent/api/pm/base.py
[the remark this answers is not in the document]
geokala: So CI is running a test that is significantly different from general use? That's not too good. Perhaps we should stop doing that and rely on the inte/sys tests more?

cloudify_agent/installer/script.py
geokala: But you defined install_script earlier in the commit (or left it defined) but then just overrode it here- unless I misread?

tehasdf: nah, there's the `install_script` method on the `AgentInstallationScriptBuilder`, but this is a module-level function (which eventually calls that method)

cloudify_agent/api/pm/base.py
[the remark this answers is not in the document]
tehasdf: Eventually so, but I think I've already removed enough tests in this PR :)

geokala (approved):

PR #875, opened 2023-03-10
RD-7005 Put the resources/ dir under agent dir

tehasdf:
Using a global location (/tmp/resources) by default a bad idea because:
- it'll be accessible by other agents
- the agent user doesn't necessarily have write permissions there


tehasdf:
counterpart for master: #874

mateumann (approved):

PR #881, opened 2023-03-14
RD-205 windows package: provide a stub "venv"

tehasdf:
We can't use venv directly, so let's use virtualenv, and pretend that it's venv. Heh.


tehasdf:
This reverts #880

